PR #35, opened 2020-05-16
Add deno.unstable configuration

Qu4k:
* Added `deno.unstable` as configuration option
* Changed extension name to `deno` and publisher to `denoland`

## Motivation
A user might not want to support unstable Deno features and should have the option to choose if enable unstable type definition

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
-  "name": "vscode_deno",
+  "name": "deno",
   "displayName": "%deno.displayName%",
   "description": "%deno.description%",
   "version": "1.24.0",
-  "publisher": "the deno authors",
+  "publisher": "denoland",
   "icon": "deno.png",
   "galleryBanner": {
     "color": "#3B3738",
@@ -106,6 +106,11 @@
           ],
           "default": "terse",
           "description": "Enables logging of the Deno server to a file. This log can be used to diagnose Deno Server issues. The log may contain file paths, source code, and other potentially sensitive information from your project."
+        },
+        "deno.unstable": {
+          "type": "boolean",
+          "default": true,
+          "description": "%deno.config.unstable%"
         }
       }
     },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "vscode_deno",
+  "name": "deno",
```
bartlomieju, line 2: revert

package.json
```diff
@@ -1,9 +1,9 @@
...
   "displayName": "%deno.displayName%",
   "description": "%deno.description%",
   "version": "1.24.0",
-  "publisher": "the deno authors",
+  "publisher": "denoland",
```
bartlomieju, line 6: revert

package.json
Qu4k: I was not able to get the extension working locally with `name` set to `vscode_deno`, possibly because of hardcoded value in [extension.ts](https://github.com/denoland/vscode_deno/blob/1a6b9060ffe9a0d5d6ba77dfb3e9399c88f5aa06/client/src/extension.ts#L23) 
```typescript
const denoExtensionId = "denoland.deno";
```

same goes for `publisher` set to `the deno authors`

package.json
Qu4k: https://github.com/denoland/vscode_deno/pull/35#discussion_r426184853

justjavac: the extension has not been published to the vscode market place with an official account, so we will keep this unchanged. When pr is merged into master, I will manually publish to justjavac.vscode-deno. 

When the extension is stable, I will mark justjavac.vscode-deno as deprecated and guide the user to install the official extension

Qu4k: Ok, thanks for the explanation. I will revert my changes

Qu4k:
I reverted my changes on package.json

PR #84, opened 2020-06-11
revert changes to get it working again

ry:
Just trying to get back to a good state quickly. I don't have time to dig into the extension right now and properly fix it.


jsejcksn:
> Just trying to get back to a good state quickly.

What does this mean, and what does it reference?

Issue #87, opened 2020-06-14
Disable extension for a subdirectory


taigah:
I know how to disable the extension in a project by using a `.vscode/settings.json` file but it does not seem to work for subdirectories of a project.

My problem is that I have a project in which there is deno code and browser javascript but because of the deno extension I have lint errors like "Could not find name 'alert'".

How can I prevent that?

adekau:
What I've been doing as a workaround is having two VSCode windows open, one with my backend code folder opened and one with my frontend code folder opened. Then using the commands for enabling and disabling Deno from the command palette puts a .vscode folder in each so you can control them separately.

Flaque:
We're using Deno inside a monorepo, so I'm going to try and play around with building a config option for this. No guarantees or anything, first time doing a VS Code extension, so don't let me stop anyone else from taking this.

## Proposal 

Deno VS Code should support a `deno.paths` variable that lets you specify folders or files via blob syntax as deno enabled.

For example, a config like this:
```
{
  "deno.paths": ["packages/deno-project/**", "*.deno.ts"]
}
```

Would set anything in the `packages/deno-project` folder as well as any file like `foo.deno.ts`.

lucacasonato:
Closing in favor of https://github.com/denoland/vscode_deno/issues/314

PR #92, opened 2020-06-19
Add "enablePatterns" config option

Flaque:
# Overview

This PR adds a `deno.enablePatterns` config option that lets you specify particular paths or folders where the extension is enabled. This is useful for projects that contain both Node code and Deno code, such as the `vscode_deno` repo itself. This relates to #87 

Example usage:
```json
{
  "deno.enablePatterns": ["packages/deno/**", "*.deno.ts"]
}
```

Each item in the array matches a regex and only one item needs to match to enable the extension on that file. 

## Disclaimer

This is more of a suggestion. Totally understand if this is the wrong naming convention, or the wrong way to go about this. 

## Testing Done

I've played around with this in a number of different configurations in [this example repo I setup](https://github.com/Flaque/deno_ext_example). Though it might be a good idea for someone else to play around with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Note the VS Code based configuration overrides the `tsconfig` configuration.
 
 - `deno.enable` - Enable/disable this extension. Default is `true`.
 
+- `deno.enablePatterns` - An array of regexes that matches files Deno should be enabled on. Default is `["*"]` (matches all files). Paths are relative to the workspaces directory, so for example `["packages/"']` will look for the `packages` folder in your project.
+
 - `deno.alwaysShowStatus` - Always show the Deno status bar item. Default is `true`.
 
 - `deno.importmap` - The Path of import maps. Default is `null`.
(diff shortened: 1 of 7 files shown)


jsejcksn:
Some questions to consider:

- Is the intended use for this to match patterns relative to the workspace directory or root directory?
- Regex patterns and/or glob patterns?
- Is `deno.enablePatterns` a more suitable name for an array of patterns?

Flaque:
@jsejcksn 

> Is the intended use for this to match patterns relative to the workspace directory or root directory?

Workspace directory, I'll add a note in the docs. 

> Regex patterns and/or glob patterns?

Regex, and same.

> enablePatterns

Agreed, that's a better name, I'll update.

David-Else:
@Flaque This looks amazing! I am trying to use Deno just for testing along with Snowpack for building. I need the `test` directory to be just for Deno test code with it's `.ts` extensions. The latest update on that seems to be https://github.com/microsoft/TypeScript/issues/37582

Who has the authority to merge this, it has been nearly a month. Do you think it is ready now?

Flaque:
🎉🎉🎉🎉

PR #103, opened 2020-07-06
Merge `typescript-deno-plugin` into `vscode_deno`

### LINKED ISSUE #78: [Discussion] Possible solution to fix delays and versioning confusion: Merge `typescript-deno-plugin` into `vscode_deno`?
# Versioning confusion

At the moment we have:

The `vscode_deno 1.24.0` official plugin that depends on, and is currently bundled with the 3rd party `typescript-deno-plugin 1.29`

- `typescript-deno-plugin v1.29.0` was released 19 days ago
-  It is now up to version `v1.31.0` which was released 8 days ago. `vscode_deno 1.24.0` still includes `typescript-deno-plugin 1.29`, not the new one.

There is no changelog for `vscode_deno` https://github.com/denoland/vscode_deno/releases

And the VS Marketplace one does not make sense: https://marketplace.visualstudio.com/items/denoland.vscode-deno/changelog

> Moved from https://github.com/justjavac/vscode-deno to https://github.com/denoland/vscode_deno in order to have an "official" Deno plugin. Thank you 迷渡!

# Problems of having a split codebase
It seems half the bug fixes are more relevant to the `typescript-deno-plugin`, which is not official and is not where users are directed to for reporting issues. If they do get through to that repo, and get the bug fixed, then they don't get merged into the `vscode_deno` extension right away. There is a random time-lapse that means all the error reports on the two different repos are out of sync.

I don't want to sound like I am complaining, and really appreciate the work @justjavac  and others are doing, but something seems fundamentally wrong with the way this extension is setup and operates.

There should be one central place that people report bugs, and at least a few different people with the authority to accept the pull requests for the official Deno plugin as a whole, not split up. Having the extension split over two separate codebases, one official and the other 3rd party seems very complicated and unnecessary.

I spent a while this morning puzzling over why my JS was getting treated like it was TS even with `"deno.enable": false`. When I turn off the extension then it works! I think this bug was fixed in `typescript-deno-plugin v1.30.0`, 9 days ago https://github.com/denoland/vscode_deno/issues/60,  but I don't know as it is not yet merged into `vscode_deno`

I hope this issue has not been too confused, but I am very confused by the whole VS Code Deno extension situation.

THANKS again to everyone involved!

justjavac:
fixed #78


jsejcksn:
What is blocking this from being merged?

mattgaspar:
@justjavac I think there is a problem with the deno.disabled setting not getting applied.  When the setting is true/checked, the Deno Language Service tries to keep running and throws errors.  This also occurs with the published version of the extension.
But the auto formatting bug seems to be fixed.

image

ry:
Old PR. We merged @Axetroy's version of typescript-deno-plugin.

Issue #114, opened 2020-07-22
Add some way to quickly import std modules


CGQAQ:
For example: after I type `https://deno.land/std@`, give me completions of all std lib versions out there; and when I select a version and hit tab, give me completions of all std module in [here](https://deno.land/std)

And of course, add a snippet code for example `importstd`,  hit tab key give me  `import {<2. hit tab input here>} from 'https://deno.land/std@<1. input here>'`

CGQAQ:
cc @ry

PR #115, opened 2020-07-22
add 2 snippets: impstd & importstd and import std module url completion

CGQAQ:
$0: denotes the final cursor position
$1..n: tabstops in sequence

impstd =>  `import {$3} from 'https://deno/std@$1/$2';$0"`
importstd => `import (foo|{foo}| * as foo) from 'https://deno/std@$1/$2';$0"`

- [x]  fetch std library information dynamically  from github
- [x]  import completion
- [x]  std version completion
- [x]  cache the result
- [x]  add command to clear the cache to allow user to refresh local import std cache

## Screenshots: 
image
image
image
image
image

Implementation of #114

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -2,6 +2,10 @@
 
 import * as vscode from "vscode";
 import * as lsp from "vscode-languageclient";
+import {
+  ExecuteCommandRequest,
+  RevealOutputChannelOn,
+} from "vscode-languageclient";
 
 /**
  * Represent a vscode command with an ID and an impl function `execute`.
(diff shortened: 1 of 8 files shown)


lucacasonato:
@CGQAQ Thanks, this is a cool feature. Since you implemented this a lot in this repository has changed, and also there has been a complete rewrite of the deno.land/x registry powering `deno.land/std`, so this PR will need some changes. If you want to continue that is cool, otherwise I can pick up this PR.

I think this feature could be extended to support all of `deno.land/x`, not just `std` with these new API endpoints on the registry:
- `https://cdn.deno.land/$MODULE/meta/versions.json` to get a list of versions for a module
- `https://cdn.deno.land/$MODULE/versions/$VERSION/meta/meta.json` to get a file tree of the module
- `https://api.deno.land/modules?limit=10&query=$QUERY` to search `deno.land/x`

Also I think we should limit our cache to a max of 1 hour, after which we would lazily fetch new versions automatically.

CGQAQ:
> @CGQAQ Thanks, this is a cool feature. Since you implemented this a lot in this repository has changed, and also there has been a complete rewrite of the deno.land/x registry powering `deno.land/std`, so this PR will need some changes. If you want to continue that is cool, otherwise I can pick up this PR and finish it.
> 
> I think this feature could be extended to support all of `deno.land/x`, not just `std` with these new API endpoints on the registry:
> 
> * `https://cdn.deno.land/$MODULE/meta/versions.json` to get a list of versions for a module
> * `https://cdn.deno.land/$MODULE/versions/$VERSION/meta/meta.json` to get a file tree of the module
> * `https://api.deno.land/modules?limit=10&query=$QUERY` to search `deno.land/x`
> 
> Also I think we should limit our cache to a max of 1 hour, after which we would lazily fetch new versions automatically.

OK, I'll change it, thank you for your information

PR #117, opened 2020-07-25
re-added deno fmt on save

JohnstonCode:
A First pass implementation of getting `fmtOnSave` to work again. Not 100% sure how to handle any errors from the `execa` call.


JohnstonCode:
I see that it is available through the LSP and vscode formatting.

PR #125, opened 2020-08-07
merge axetroy's extension

bartlomieju:
CC @ry @axetroy @justjavac

PR #126, opened 2020-08-10
Include deno.enablePatterns on config schema

turadg:
A new config option, `deno.enablePatterns`, was added in #92. The [config schema](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) wasn't added, probably because there's no UI editor for "array" type. However, it's still worth defining.

1. To avoid this confusing message:
Screen Shot 2020-08-09 at 5 31 18 PM

2. To get an indication in the Settings UI that the key exists and can be edited in the JSON:
> Other types, such as object and array, aren't exposed directly in the settings UI, and can only be modified by editing the JSON directly. Instead of controls for editing them, users will see a link to Edit in settings.json as shown in the screenshot above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,10 @@
           "default": false,
           "description": "%deno.config.enabled%"
         },
+        "deno.enablePatterns": {
+          "type": ["array"],
+          "description": "%deno.config.enablePatterns%"
+        },
         "deno.alwaysShowStatus": {
           "type": "boolean",
           "default": true,


ry:
Please merge with master

David-Else:
@turadg @ry I think https://github.com/denoland/vscode_deno/pull/92 was wiped out in the recent merge so this PR is now invalid.

turadg:
@David-Else thanks for pointing out that #125 wiped out the `enablePatterns` feature @Flaque created in #92.

That feature is essential in codebases that mix Deno and non-Deno code.

@bartlomieju would you accept it again?

bartlomieju:
@turadg sure! Sorry I didn't carry it over

PR #153, opened 2020-08-21
reimplement deno.enablePatterns feature

turadg:
Attempting to add https://github.com/denoland/vscode_deno/pull/92/ after it was [removed from master](https://github.com/denoland/vscode_deno/pull/126#issuecomment-674508634). 

However, in my testing it's not working as expected. Even documents that don't match are getting Deno errors when `enabled` is true. For example,
Screen Shot 2020-08-21 at 2 48 06 PM

@Flaque did #92 have this problem?

---
This change is [Reviewable](https://reviewable.io/reviews/denoland/vscode_deno/153)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ const M = array.getMonoid<number>();
 console.log("concat Array", M.concat([1, 2], [2, 3]));
 ```
 
+- `deno.enablePatterns` - An array of regexes that matches files Deno should be enabled on. Default is `["*"]` (matches all files). Paths are relative to the workspaces directory, so for example `["packages/"']` will look for the `packages` folder in your project.
+
 </details>
 
 ## Usage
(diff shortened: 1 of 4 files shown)


lucacasonato:
I don't think we should add our own language ID. We don't change the syntax at all (`typescriptreact` does because of JSX). We only change semantics.

turadg:
> We don't change the syntax at all

Deno has its own standard formatting which is different from other TypeScript formatters like Prettier.

> We only change semantics.

What's a better way to determine whether a file is a Deno .ts file or a non-Deno .ts file?

> I don't think we should add our own language ID.

What are the downsides of adding a language ID? I believe it could inherit from `typescript` such that it wouldn't diverge except for the specific parts necessary.

lucacasonato:
@axetroy what do you think?

axetroy:
> @axetroy what do you think?

I didn't try this before.

Add a new file type association `typescriptdeno`. I don't know if it will affect the original typescript function.

lucacasonato:

PR #166, opened 2020-08-29
Problem: (Fix #47) Deno executable path and DENO_DIR are not configurable

### LINKED ISSUE #47: Configure deno executable path
I use `direnv` to set PATH per project. Which means:
```
~$ which deno
deno not found

~$ cd deno-playground 
direnv: loading .envrc                                                                                                                                                           
direnv: export ~PATH

~/deno-playground$ which deno
/home/user/.deno-1.0/bin/deno
```

I think a settings field like "deno.denoPath" should be introduced just like "python.pythonPath".

calvinlauco:
Solution: Add configurable path to VSCode plugin settings

image

### DIFF
--- a/core/deno.ts
+++ b/core/deno.ts
@@ -7,7 +7,7 @@ export function getDenoDir(): string {
   // On Windows: %LOCALAPPDATA%/deno (%LOCALAPPDATA% = FOLDERID_LocalAppData)
   // On macOS: $HOME/Library/Caches/deno
   // If something fails, it falls back to $HOME/.deno
-  let denoDir = process.env.DENO_DIR;
+  let denoDir = process.env.DENO_DIR || process.env.VSCODE_CUSTOM_DENO_DIR;
   if (denoDir === undefined) {
     switch (process.platform) {
       /* istanbul ignore next */
--- a/core/deno.ts
+++ b/core/deno.ts
@@ -7,7 +7,7 @@ export function getDenoDir(): string {
   // On Windows: %LOCALAPPDATA%/deno (%LOCALAPPDATA% = FOLDERID_LocalAppData)
   // On macOS: $HOME/Library/Caches/deno
   // If something fails, it falls back to $HOME/.deno
-  let denoDir = process.env.DENO_DIR || process.env.VSCODE_CUSTOM_DENO_DIR;
+  let denoDir = process.env.VSCODE_CUSTOM_DENO_DIR || process.env.DENO_DIR;
   if (denoDir === undefined) {
     switch (process.platform) {
       /* istanbul ignore next */
--- a/server/src/deno.ts
+++ b/server/src/deno.ts
@@ -90,20 +90,17 @@ class Deno {
     return formattedCode;
   }
   private async getExecutablePath(): Promise<string | undefined> {
-    let denoPath: string | undefined;
+    const denoPath = process.env.VSCODE_DENO_EXECUTABLE_PATH;
+    if (denoPath !== undefined) {
+      return denoPath;
+    }
 
     const denoRoot = process.env.DENO_INSTALL_ROOT || process.env.DENO_INSTALL;
-    if (denoRoot === undefined) {
-      denoPath = process.env.VSCODE_DENO_EXECUTABLE_PATH;
-    } else {
-      denoPath = path.join(denoRoot, "deno");
+    if (denoRoot !== undefined) {
+      return path.join(denoRoot, "deno");
     }
 
-    if (denoPath === undefined) {
-      denoPath = await which("deno").catch(() => Promise.resolve(undefined));
-    }
-
-    return denoPath;
+    return which("deno").catch(() => Promise.resolve(undefined));
   }
   private async getDenoVersion(): Promise<Version | undefined> {
     const { stdout, stderr } = await execa(this.executablePath as string, [
--- a/server/src/deno.ts
+++ b/server/src/deno.ts
@@ -133,7 +133,6 @@ class Deno {
     }
   }
   private async getDenoVersion(): Promise<Version | undefined> {
-    console.log;
     const { stdout, stderr } = await execa(this.executablePath as string, [
       "eval",
       "console.log(JSON.stringify(Deno.version))",
--- a/server/src/deno.ts
+++ b/server/src/deno.ts
@@ -5,6 +5,8 @@ import execa from "execa";
 import which from "which";
 import * as semver from "semver";
 
+import { pathExists } from "../../core/util";
+
 type Version = {
   deno: string;
   v8: string;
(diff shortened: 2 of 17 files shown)


core/deno.ts
```diff
@@ -7,7 +7,7 @@ export function getDenoDir(): string {
   // On Windows: %LOCALAPPDATA%/deno (%LOCALAPPDATA% = FOLDERID_LocalAppData)
   // On macOS: $HOME/Library/Caches/deno
   // If something fails, it falls back to $HOME/.deno
-  let denoDir = process.env.DENO_DIR;
+  let denoDir = process.env.DENO_DIR || process.env.VSCODE_CUSTOM_DENO_DIR;
```
axetroy, line 10: `process.env.VSCODE_CUSTOM_DENO_DIR` should be a higher priority

```ts
let denoDir = process.env.VSCODE_CUSTOM_DENO_DIR || process.env.DENO_DIR;
```

axetroy (changes requested):

core/deno.ts
calvinlauco: Ok, I have updated both `DENO_DIR` and executable path settings to have a higher priority.

core/deno.ts
```diff
@@ -7,7 +7,7 @@ export function getDenoDir(): string {
   // On Windows: %LOCALAPPDATA%/deno (%LOCALAPPDATA% = FOLDERID_LocalAppData)
   // On macOS: $HOME/Library/Caches/deno
   // If something fails, it falls back to $HOME/.deno
-  let denoDir = process.env.DENO_DIR;
+  let denoDir = process.env.VSCODE_CUSTOM_DENO_DIR || process.env.DENO_DIR;
```
axetroy, line 10: One more thing.

Because the functions in the core will be used in client/server/typescript-deno-plugin

So when the configuration changes, it cannot be changed immediately, and the extension needs to be restarted.

To fix this, We can watch configuration changes and then set environment variables

axetroy (changes requested):

core/deno.ts
calvinlauco: Thanks @axetroy .

Yes, I have the line `this.startDenoLangaugeServer()` in the `sync` function at https://github.com/denoland/vscode_deno/pull/166/files#diff-bc3c0f2fe3aca9ddbe96347ffb00641eR409-R433

and the `sync` function is "registered" to watch for configuration https://github.com/denoland/vscode_deno/pull/166/files#diff-bc3c0f2fe3aca9ddbe96347ffb00641eR615-R617

so it will watch for configuration changes and restart the server already.

calvinlauco:
I have just added the valid check for the `deno_executable_path` setting.

typescript-deno-plugin/src/plugin.ts
```diff
@@ -224,12 +224,7 @@ export class DenoPlugin implements ts_module.server.PluginModule {
         }
 
         const importMapsFilepath = this.configurationManager.config.import_map
-          ? path.isAbsolute(this.configurationManager.config.import_map)
-            ? this.configurationManager.config.import_map
-            : path.resolve(
-                project.getCurrentDirectory(),
-                this.configurationManager.config.import_map
-              )
+          ? this.configurationManager.config.import_map
```
axetroy, line 227: Missing the config update in typescript-deno-plugin


https://github.com/denoland/vscode_deno/blob/4e23f8c7e3922d360eb6fa75c885c34c4538a250/typescript-deno-plugin/src/plugin.ts#L557-L559

axetroy (changes requested):

core/deno.ts
axetroy: > so it will watch for configuration changes and restart the server already.

Why doesn't the LSP watch configuration changes and then update the environment variables?

This should be a simple model.

Restarting the LSP is not necessary

calvinlauco: Do you mean to use some pub sub channel (e.g. LangeuageServer `sendNotification()` + `onNotification()`) to notify the running server of the environment changes so that I don't need to restart it?

p.s. I haven't read through the whole implementation of the language server yet. Some guidance to enlighten me is greatly appreciated. Thank you.

calvinlauco: Oh, I re-read what you said. Do you mean there is an even simpler way to watch for VSCode configuration changes inside the DenoLangaugeServer instance? So it does not even need to be triggered from the extension side on `watchConfiguration()` callback?

calvinlauco: Oh, I re-read what you said. Do you mean there is an even simpler way to watch for VSCode configuration changes inside the DenoLangaugeServer instance? So it does not even need to be triggered from the extension side on `watchConfiguration()` callback?

calvinlauco: @axetroy Thanks for the advice.

I have updated to listen to the config changes inside the language server.

typescript-deno-plugin/src/plugin.ts
calvinlauco: @axetroy I checked the file and it seems that the only place that reads `DENO_DIR` is https://github.com/denoland/vscode_deno/pull/166/files#diff-ea9fa7070d48e644600141672d50fb9cR189-R207, so I have added the config update there. Please correct me if I am wrong.

server/src/deno.ts
```diff
@@ -89,13 +92,48 @@ class Deno {
...
+        return false;
+      }
+
+      return !!JSON.parse(stdout).deno;
+    } catch (err) {
+      return false;
+    }
   }
   private async getDenoVersion(): Promise<Version | undefined> {
+    console.log;
```
axetroy, line 136: redundant code

axetroy (changes requested):

server/src/deno.ts
calvinlauco: Code updated. Thanks!

lucacasonato:
@axetroy please re-review if you have some time

typescript-deno-plugin/src/plugin.ts
```diff
@@ -182,6 +186,22 @@ export class DenoPlugin implements typescript.server.PluginModule {
       }
 
       // Get typescript declaration File
+      if (
```
axetroy, line 189: This seems to only run when the project is initialized.

So if the configuration changes, it doesn’t seem to be notified.

I think it might be better to change env in here

https://github.com/denoland/vscode_deno/blob/65a0928ef9a5db1640ce2bd95f899a8c72ba2ff7/typescript-deno-plugin/src/plugin.ts#L557-L559

https://github.com/denoland/vscode_deno/blob/65a0928ef9a5db1640ce2bd95f899a8c72ba2ff7/typescript-deno-plugin/src/plugin.ts#L546-L552

axetroy (changes requested):

lucacasonato:

PR #173, opened 2020-09-01
Add support for URLs with non-default ports

### LINKED ISSUE #171: Cached remote modules with non-default ports are not resolved
See https://github.com/denoland/vscode_deno/issues/24#issuecomment-649918614 for details.

Before the project switched from @justjavac's code to @axetroy's code, there was a bug where the extension did not resolve module URLs with non-default ports. This issue persists after the extension switched codebases. Previously, I tracked down the bug to `typescript-deno-plugin` and [patched it](https://github.com/justjavac/typescript-deno-plugin/pull/63), but I'm guessing it's not going to be merged into this extension now, so I'm filing this issue.

**Versions**

vscode: `1.48.2`
deno: `1.3.2`
extension: `2.0.16`

jsejcksn:
Ref. Deno source: https://github.com/denoland/deno/blob/2da084058397efd6f517ba98c9882760ec0a7bd6/cli/disk_cache.rs#L55

Ref. previous PR: https://github.com/justjavac/typescript-deno-plugin/pull/63

This adds support for port numbers in remote module resolution. I didn't see any places that would support a test for this, so please test with a local file server before merging.

I think this should resolve #171.

### DIFF
--- a/core/deno_cache.ts
+++ b/core/deno_cache.ts
@@ -125,7 +125,7 @@ export class CacheModule implements DenoCacheModule {
       targetOriginDir = path.join(
         getDenoDepsDir(),
         url.protocol.replace(/:$/, ""), // https: -> https
-        url.hostname + (url.port ? `_PORT${url.port}` : url.port)
+        `${url.hostname}${url.port ? `_PORT${url.port}` : ""}`
       );
     }
     // invalid
--- a/core/deno_cache.ts
+++ b/core/deno_cache.ts
@@ -125,7 +125,7 @@ export class CacheModule implements DenoCacheModule {
       targetOriginDir = path.join(
         getDenoDepsDir(),
         url.protocol.replace(/:$/, ""), // https: -> https
-        url.hostname
+        url.hostname + (url.port ? `_PORT${url.port}` : url.port)
       );
     }
     // invalid
--- a/core/module_resolver.ts
+++ b/core/module_resolver.ts
@@ -75,7 +75,7 @@ export class ModuleResolver implements ModuleResolverInterface {
     const originDir = path.join(
       getDenoDepsDir(),
       url.protocol.replace(/:$/, ""), // https: -> https
-      url.hostname
+      `${url.hostname}${url.port ? `_PORT${url.port}` : ""}`, // hostname.xyz:3000 -> hostname.xyz_PORT3000
     );
 
     const hash = hashURL(url);
--- a/core/module_resolver.ts
+++ b/core/module_resolver.ts
@@ -75,7 +75,7 @@ export class ModuleResolver implements ModuleResolverInterface {
     const originDir = path.join(
       getDenoDepsDir(),
       url.protocol.replace(/:$/, ""), // https: -> https
-      `${url.hostname}${url.port ? `_PORT${url.port}` : ""}`, // hostname.xyz:3000 -> hostname.xyz_PORT3000
+      `${url.hostname}${url.port ? `_PORT${url.port}` : ""}` // hostname.xyz:3000 -> hostname.xyz_PORT3000
     );
 
     const hash = hashURL(url);
--- a/core/module_resolver.test.ts
+++ b/core/module_resolver.test.ts
@@ -28,6 +28,8 @@ test("core / module_resolver: resolve module from Deno cache", () => {
       "./sub/mod.ts",
       "/esm/mod.ts",
       "https://example.com/esm/mod.ts",
+      "https://example.com:443/esm/mod.ts",
+      "http://localhost:3000/mod.ts",
       "./module_not_exist.ts",
       "https://module.not.exist.com/mod.ts",
     ])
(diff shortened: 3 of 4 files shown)


core/module_resolver.ts
```diff
@@ -75,7 +75,7 @@ export class ModuleResolver implements ModuleResolverInterface {
     const originDir = path.join(
       getDenoDepsDir(),
       url.protocol.replace(/:$/, ""), // https: -> https
-      url.hostname
+      `${url.hostname}${url.port ? `_PORT${url.port}` : ""}` // hostname.xyz:3000 -> hostname.xyz_PORT3000
```
axetroy, line 78: could you add a test for this?

axetroy (changes requested):

core/module_resolver.ts
axetroy: ------------------------------------|---------|----------|---------|---------|-------------------
File                                | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
------------------------------------|---------|----------|---------|---------|-------------------
All files                           |     100 |    99.59 |     100 |     100 |                   
 cache.ts                           |     100 |      100 |     100 |     100 |                   
 cache_map.ts                       |     100 |      100 |     100 |     100 |                   
 configuration.ts                   |     100 |      100 |     100 |     100 |                   
 deno.ts                            |     100 |      100 |     100 |     100 |                   
 deno_cache.ts                      |     100 |      100 |     100 |     100 |                   
 deno_deps.ts                       |     100 |      100 |     100 |     100 |                   
 deno_normalize_import_statement.ts |     100 |      100 |     100 |     100 |                   
 deno_type_hint.ts                  |     100 |      100 |     100 |     100 |                   
 extension.ts                       |     100 |      100 |     100 |     100 |                   
 file_walker.ts                     |     100 |      100 |     100 |     100 |                   
 hash_meta.ts                       |     100 |      100 |     100 |     100 |                   
 import_map.ts                      |     100 |      100 |     100 |     100 |                   
 module_resolver.ts                 |     100 |    97.14 |     100 |     100 | 78                
 util.ts                            |     100 |      100 |     100 |     100 |                   
------------------------------------|---------|----------|---------|---------|-------------------

jsejcksn: You mean extract the behavior to a function and add a few static unit tests?

axetroy: See here

https://github.com/denoland/vscode_deno/blob/84478514f2fc3ce3e930291d0228187256334a27/core/module_resolver.test.ts#L26-L34

you need to create a folder in `./testdata/deno_dir_mainfest/deps/https/example.com_port3000`

and then create a test file

[hash]
[hash].metadata.json

Please refer to other files

jsejcksn: I took at look at the project structure.

It's not clear to me what you mean by

> and then create a test file
>
> [hash]
> [hash].metadata.json
>
> Please refer to other files

lucacasonato: @jsejcksn See `//core/testdata/deno_dir_manifest/deps`

jsejcksn: > Create a file there

I don't see any instructions for "creating a file".

By "a file" do you mean a TypeScript module file?

What is meant by "hash"? An SHA-256 hex checksum of the binary contents of the module file? Should the metadata file be named from the same checksum?

Will you be a bit more explicit, please?

lucacasonato: Sorry - here is a rough explanation:

DENO_DIR is the place where remote files are cached locally on disk. In the DENO_DIR the file paths are sha256 hex encoded to prevent issues with operating systems that don't support certain characters in file names (eg `?&=`). Next to this file there is also a file with the same sha256 hex name, that ends in `.metadata.json`.

We have this manually constructed DENO_DIR in `//core/testdata` so the tests have something to resolve.

So to create a file:
1. Take the file path and sha256 hash it, and use hex representation.
2. Create file with this hash as a name in the folder with the correct domain name (eg `anotherport.example.com_PORT3000` if the domain name is `anotherport.example.com:3000`. In this file the regular file contents go.
3. Create a metadata file (name is `[hash].metadata.json` in the same directory. This should contain a JSON object that looks something like this:
```json
{
  "headers": {},
  "url": "https://anotherport.example.com:3000/path/mod.ts"
}
```

Then you need to add a test for this file here: https://github.com/denoland/vscode_deno/blob/84478514f2fc3ce3e930291d0228187256334a27/core/module_resolver.test.ts#L26-L34

jsejcksn: Thanks. That gives me a decent understanding now of what was requested, and I think I can update the test based on that information.

jsejcksn: It looks like the test is failing, but I'm not sure what the report output means:

```
    - Expected  - 5
    + Received  + 1
```

https://github.com/denoland/vscode_deno/pull/173/checks?check_run_id=1071531498#step:11:14

lucacasonato: I don't know if the extension even supports `http` schemed files. I'll try it now.

lucacasonato: So it does seem to support http scheme. I'll investigate some more...

lucacasonato: @jsejcksn Found it - there was another place where the port was not being joined to the hostname (/core/deno_cache.ts:128). I have pushed a fix to your branch.

core/deno_cache.ts
```diff
@@ -125,7 +125,7 @@ export class CacheModule implements DenoCacheModule {
       targetOriginDir = path.join(
         getDenoDepsDir(),
         url.protocol.replace(/:$/, ""), // https: -> https
-        url.hostname
+        url.hostname + (url.port ? `_PORT${url.port}` : url.port)
```
jsejcksn, line 128:
```suggestion
        `${url.hostname}${url.port ? `_PORT${url.port}` : ""}`
```

@lucacasonato Thanks for finding this!

Also, not sure if the Deno linter has a rule about using the addition operator for string concatenation, but in any case, I think consistency is good for this line and https://github.com/denoland/vscode_deno/pull/173/files#diff-f5bbeb1e4c5f2facaa107bf559832939R78

jsejcksn:
It looks like the test is passing now. Let me know if any other changes are needed. Thanks for both the info and help, @lucacasonato!

lucacasonato (approved):
LGTM - thanks Jesse!

PR #181, opened 2020-09-05
Try to move import enhancement to server for better performance

CGQAQ:
### SHOWCASE
import_enhancement

import_enhancement 2

import_enhancement 3


CGQAQ:

lucacasonato:
From my initial testing the `std` completion for `import {} from "https://deno.land/"`, completes `import {} from "https://deno.land/std/"` instead of `import {} from "https://deno.land/std@"`. Could you change this?

CGQAQ:
> From my initial testing the `std` completion for `import {} from "https://deno.land/"`, completes `import {} from "https://deno.land/std/"` instead of `import {} from "https://deno.land/std@"`. Could you change this?

The std default to not need a version, should I just remove the trailing / ?

lucacasonato:
I think it should default to using versioned imports, and if you give the @, people are more likely to actually set a version explicitly.

CGQAQ:
> I think it should default to using versioned imports, and if you give the @, people are more likely to actually set a version explicitly.

import_enhancement 4

lucacasonato:
Thanks. @axetroy do you have any review comments?

core/permcache.ts
```diff
@@ -0,0 +1,229 @@
+// Permanent version of the Cache from cache.ts in the same folder
+import { promises as fsp } from "fs";
+import fs from "fs";
+import path from "path";
+
+export function getVSCodeDenoDir(): string {
```
axetroy, line 6: I am a little curious, why not use the system cache folder? Or use the vscode_deno extension folder

axetroy (changes requested):

core/permcache.ts
CGQAQ: how to get extension folder?

CGQAQ: image
But you can change it by setting this environment variable

axetroy: @CGQAQ you can get it via `context.extensionPath` in client-side.

get it via env `VSCODE_DENO_EXTENSION_PATH` in server side

https://github.com/denoland/vscode_deno/blob/65a0928ef9a5db1640ce2bd95f899a8c72ba2ff7/client/src/extension.ts#L201-L206

this directory will change every time when the extension is updated

eg.

denoland.vscode_deno-1.0.0
denoland.vscode_deno-1.1.0

CGQAQ: image

So, this will be decent?

CGQAQ:
@axetroy Any others need to change?

axetroy (approved):

lucacasonato (approved):
Thanks @CGQAQ for implementing this, and thanks @axetroy for the review

PR #191, opened 2020-09-08
Fix #190

### LINKED ISSUE #190: Deno x module  list cache will always recache everytime when open vscode after the first cache expired
Deno x module  list cache will always recache every time you open vscode after the first cache expired

CGQAQ:
Fixed #190


CGQAQ:

lucacasonato:
LGTM

PR #192, opened 2020-09-08
Fix #188

### LINKED ISSUE #188: bug: 'ignore for this line' quickfix does not work on line 1
**Describe the bug**

If you try to use the `ingore 'rule' for this line` quickfix for the `deno lint` integration on the first line, this will fail with error `Illegal value for 'line'`.

**To Reproduce**

Create file:
```ts
for (;;) {}
```
Try to use the `ignore 'no-empty' for this line` quickfix on line 1.

**Expected behavior**

This should work.

**Versions**

vscode: 1.48.2
deno: 1.3.3
extension: 2.2.0

cc @axetroy

CGQAQ:
Fixed #188

Plus

Fixed a small issue
when currentline have trailing space e.g.: 
```js
   for(;;){}   
will result
      // deno-lint-ignore no-empty
   for(;;){}
```


CGQAQ:

lucacasonato (approved):
Thanks @CGQAQ

PR #196, opened 2020-09-09
Add include and exclude

CGQAQ:
Add include and exclude config

pr

Part of roadmap 2 (#194)

Closes #75, Closes  #50, Closes #39, Closes #60, Closes #87

Issue #194, opened 2020-09-08
Roadmap 2


lucacasonato:
Now that most burning fires have been put out #121 can be closed and it is time for a new roadmap. The goals for the next few weeks and months should be stabilizing the extension, improving compatibility with mixed Node + Deno projects, and making the extension more easy to use. Here are some ideas in rough order of priority (top is most important):

- [x] `editor.defaultFormatter` should not be hijacked by `deno fmt`
- [ ] ~Implement `"deno.include"` setting to enable the extension on files matching this path.~ not possible until `deno lsp`
- [ ] ~Implement `"deno.exclude"` setting to disable the extension on files matching this path.~ not possible until `deno lsp`
- [ ] Various improvements to language server. (#187)
- [x] Add Deno debugging configuration.
- [x] Re-add `Enable Deno` and `Disable Deno` commands to quickly switch `"deno.enable"` in `.vscode/settings.json`. (#170)
    - This was instead covered by `Init` command.
- [ ] Support re-exports from files like `deps.ts` for autocomplete.
- [ ] Support configuration in workspace and user settings, not just project settings.

And various high priority bugs that need to be fixed:
- [ ] Syntax highlighting is somewhat broken (#150)
- [ ] `@deno-types` does not work on first load (#97)

---

The future goal is still to integrate LSP into the `deno` binary.

brandonkal:
Nice. Also
#124
#113
are still bugs.
-1 on a toggle command. There are many existing generic settings toggle solutions so no need to reimplement the wheel there.

I'd also love a deno.enable auto where a file is recognized as Deno if it contains the string "Deno" (either as namespace usage or in a comment).

CGQAQ:
@lucacasonato Please create an issue for each item in the checklist, so I can reference in pr

callionica:
If console.log isn't writing to the VS Code debugger output pane, it's hard to see the VS Code integration as working, so it'd be great to get the fix to Deno #4502

https://github.com/denoland/deno/issues/4502

Maybe this is covered by the debugging item?

nayeemrmn:
@callionica That's independent of VSCode integration, since the same issue will occur with Chrome devtools for instance.

callionica:
@nayeemrmn Yeah, I recognise that it's under a different repo. I hope this roadmap is being used to show the path to completing specific customer scenarios and fulfilling user expectations. That's a good way to prioritise where to put your resources.

[...]

PR #197, opened 2020-09-09
Add enable disable command

### LINKED ISSUE #170: [regression] deno enable/disable command palette entries removed
There is no longer a VS Code command palette option to "Enable Deno" or to "Disable Deno" (it seems I must manually create the workspace settings file in every new workspace).

**Screenshots**

vscode-command-palette-deno

**Versions**

vscode: `1.48.2`
deno: `1.3.2`
extension: `2.0.16`

CGQAQ:
Depends on #196, Please review it first

Show `enable command` when lsp is disabled, `disable command` when lsp is enabled

pr3

pr

Part of roadmap 2 (#194)

Closes #170


CGQAQ:
@bartlomieju @axetroy @lucacasonato  Ready for review

David-Else:
I remember this addition was debated before in the old extension and removed. I like the idea, but my only reservation is that now there are more options, logically they should be togglable as well? Perhaps instead `Init Deno project` creates and opens:

```
// .vscode/settings.json
{
  "deno.enable": true,
  "deno.unstable": true,
  "deno.lint": true,
}
```
This could be a nice time-saver? It would force people to make decisions on these options right away, and educate them that they exist.

CGQAQ:
I think to add every config to command is not a good idea

axetroy:
> I think to add every config to command is not a good idea

as I said. https://github.com/denoland/vscode_deno/issues/170#issuecomment-687553557

Consider adding a `deno.init` command for configuration, every field.

CGQAQ:
Don't need anymore

PR #207, opened 2020-09-13
Add init project command to quickly get things setup

CGQAQ:
pr1

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -42,6 +42,8 @@ import {
   DenoPluginConfigurationField,
 } from "../../core/configuration";
 
+import { initProject, ProjectSetting } from "./init_project";
+
 const TYPESCRIPT_EXTENSION_NAME = "vscode.typescript-language-features";
 const TYPESCRIPT_DENO_PLUGIN_ID = "typescript-deno-plugin";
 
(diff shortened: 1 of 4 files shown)


CGQAQ:
@lucacasonato sorry, mb

lucacasonato (commented):
This is looking good. Can you make the flow be like this:

1. User runs `Deno: Init` command
2. Question: `Enable --unstable?` (the user can answer `yes` or `no`)
3. If --unstable: Question: `Enable deno lint?` (the user can answer `yes` or `no`)

CGQAQ:
> If --unstable: Question: Enable deno lint? (the user can answer yes or no)

I think let user choose unstable is unnecessary, you you turn on a feature need unstable, just turn on unstable as well, otherwise just keep it off

lucacasonato:
> I think let user choose unstable is unnecessary, you you turn on a feature need unstable, just turn on unstable as well, otherwise just keep it off

Right, I just want to make users aware that there is a `unstable` option. So what about this:

1. User runs Deno: Init command
2. Question: Enable deno lint? (the user can answer yes or no)
3. Write config file and print message:
```
Deno is now set up. You can enable and disable settings (like `--unstable`) in the `.vscode/settings.json` file. Before the extension will work you need to restart VS Code.

[Restart VS Code]
```

When the user presses `[Restart VS Code]` the window should be reloaded.

CGQAQ:
@lucacasonato don't need to restart, just `reload window` is enough, I'll look for how to add a button to message

CGQAQ:
image
@lucacasonato ready for review....again

lucacasonato (approved):
LGTM - thanks

PR #209, opened 2020-09-14
refactor: import_enhanced feature now support nest.land

### LINKED ISSUE #208: Support nest land
Add [nest.land](https://nest.land/) support to import enhancement feature

CGQAQ:
pr 209

Closes #208


CGQAQ:
@lucacasonato ready for review

CGQAQ:
@lucacasonato ready for review

lucacasonato:
When autocompleting the impnest snippet I get this error:

```
[Error - 11:56:55 AM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: mods is not iterable
  Code: -32603 
```

CGQAQ:
> When autocompleting the impnest snippet I get this error:
> 
> ```
> [Error - 11:56:55 AM] Request textDocument/completion failed.
>   Message: Request textDocument/completion failed with message: mods is not iterable
>   Code: -32603 
> ```

I know what have happened, your cache at start didn't success

lucacasonato:
> I know what have happened, your cache at start didn't success

Ah - maybe we should retry to cache if the cache does not exist?

lucacasonato:
```ts
import {} from 'https://x.nest.land/std@0.69.0/';
```

Autocompleting at the end leads to:

```
[Error - 12:12:33 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'std@0.69.0' of undefined
  Code: -32603 
```

CGQAQ:
> > I know what have happened, your cache at start didn't success
> 
> Ah - maybe we should retry to cache if the cache does not exist?

pr2

CGQAQ:
> ```ts
> import {} from 'https://x.nest.land/std@0.69.0/';
> ```
> 
> Autocompleting at the end leads to:
> 
> ```
> [Error - 12:12:33 PM] Request textDocument/completion failed.
>   Message: Request textDocument/completion failed with message: Cannot read property 'std@0.69.0' of undefined
>   Code: -32603 
> ```

well that's weird

pr3

CGQAQ:
@lucacasonato Please try the latest one

CGQAQ:
> [Error - 12:12:33 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'std@0.69.0' of undefined
  Code: -32603 

@lucacasonato It is the same problem, when cache failed, the object is undefined, so it can not read using `std@0.69.0` as key from the object.

CGQAQ:
> Ah - maybe we should retry to cache if the cache does not exist?

pr 4

@lucacasonato

CGQAQ:
@lucacasonato Do you still have the issue with my last commit?

core/import_enhanced/registry.ts
```diff
@@ -0,0 +1,34 @@
...
+  // latest version should in versions[0]
+  readonly versions?: string[];
+}
+export type ModInfoList = ModInfo[];
+
+export interface Registry {
+  readonly REGISTRY_ID: string;
+  // Allow registry don't have std module mirror
+  std(): Promise<Optional<ModInfo>>;
+  stdContents(version: string[]): Promise<Optional<ModVersionMap>>;
```
lucacasonato, line 30: I think we shouldn't treat `std` like a special module. We should just use `modVersionList("std")`

lucacasonato (commented):
@CGQAQ Yup it works now. Just one minor thing

core/import_enhanced/registry.ts
CGQAQ: But If I recall correctly that in denoland, std have different api between std and x, If they're the same, just call mod one in std

kitsonk:
I am opposed to code referring to any other URL than `deno.land`. Integration should be 100% meta data, not tied to any 3rd party. And example of the meta data is suitable, but not code.

core/import_enhanced/registry.ts
lucacasonato: They are all the same API.

CGQAQ: Ok, you're right, I will change it

lucacasonato:
> I am opposed to code referring to any other URL than deno.land. Integration should be 100% meta data, not tied to any 3rd party. And example of the meta data is suitable, but not code.

@kitsonk How would you want this to work? Just the parsing of the import statement requires you to use a regex - where do we get this regex? How do we know how the URL is structured? How to we acquire this metadata? How do we know what format this metadata has? How do we know how where in the URL to suggest what information?

We could have every registry host a `/.well-known/vscodedeno.json` file with all of this information. That would likely also require that we assume that all registries have the module, version, file distinction, which is not certain. If we can come up with a good way to integrate this that doesn't limit how registries work or assume that they have multiple modules, I am all for it. That just seems like something that is not easily doable.

CGQAQ:
> I am opposed to code referring to any other URL than deno.land

The openness of the web doomed this will happen, I think deno is more like a js/ts browser on the server. A browser that don't recommend browsing some URLs other than `deno.land` is not a good idea

kitsonk:
> How would you want this to work? Just the parsing of the import statement requires you to use a regex - where do we get this regex? How do we know how the URL is structured? How to we acquire the metadata? How do we know what format this metadata has? How do we know how where in the URL to suggest what information?

@lucacasonato we need to create a specification for a repository which can be described fully by meta data.  If there are variances from that convention that should be allowed, like RegExps, they should be included in the meta data.  I haven't worked through the details, but it shouldn't require changes to extension code to add a new repository.  It should be as simple as adding something like this to the configuration:

```json
{
  "deno.registries": {
    "https://deno.land/x/": {},
    "https://x.nest.land/": {
      "latest": "https://x.nest.land/api/package/${mod}/${latest_version}"
    }
  }
}
```

If we find ourselves hard coding URLs, patterns or schemas in the code, outside of default conventional behaviour based on a provided URL, we are making an inflexible solution, let alone one that requires others to commit and change code to the plugin to enable integration.  This is something an end user should be able to trivially do of their own choice.

> A browser that doesn't recommend browsing some URLs other than deno.land is not a good idea.

Browsers shouldn't suggest what URLs people should browse.  They be default have a landing page which is owned by the browser vendor.  And browsers doesn't have hard coded URLs in its code.  You don't update your browser or your browser's dev tools code every time you want to browse a new website.

CGQAQ:
> ```json
> {
>   "deno.registries": {
>     "https://deno.land/x/": {},
>     "https://x.nest.land/": {
>       "latest": "https://x.nest.land/api/package/${mod}/${latest_version}"
>     }
>   }
> }
> ```

If we do this, we must also define what these API return format, so we can parse it

callionica:
Strongly support kitsonk’s position about not including third party URLs in Deno code for all the reasons mentioned. Want to emphasise the trust and security angle though: every third party URL shipped is a risk to Deno’s users. It wouldn’t be a consistent position to state that Deno is developed with security in mind and then link to third party websites.

CGQAQ:
Then, just forget about decentralized. Another nodejs with npm built-in is born

David-Else:
I see nothing wrong with allowing the user to add new repositories for the auto-complete, but I don't think any 3rd party ones should be built-in by default. As this extension is official, it should only officially support official resources.

Allowing the user to add things like `nest.land` is a really cool feature, but should be up to the user to configure and use at their own risk.

I think would solve the dilemma?

Ideally `vscode-deno` should define the API and vendors can choose to adopt it.

lucacasonato:
There are three paths forward:

1. Remove import enhancement feature (no import autocomplete)
2. Add out of the box support for all the registries (deno.land, nest.land, pika.dev, raw.githubusercontent.com, denopkg.com)
3. Force the user to specify all registries individually through settings (including deno.land/x)

There is no path forward where we treat deno.land/x in any special way. I approved this feature originally with the thought that more registries (like nest.land) could and would be added. Just because deno.land/x happens to be hosted on the Deno project domain does **not** make it the official recommended Deno registry. We should not actively encourage or discourage users to use a certain registry. All ecosystem features that are built into this extension should work just as well for non deno.land/x registries as they do for deno.land/x.

I am leaning towards 3 because this is the most open approach. I am going to write a proposal for this today and create an issue. We can continue discussion there.

@CGQAQ Lets wait with this PR until discussions are settled.

David-Else:
3 sounds best to me.

Soremwar:
Just casually saying, 3 also sounds like the best to me. But maybe the extension can come with a predefined list of templates for common registries?

kitsonk:
Option 3 would be best.

PR #219, opened 2020-09-18
feat: dynamic import intellisense

### LINKED ISSUE #217: discussion: multi registry import IntelliSense
We added support for IntelliSense autocomplete a week ago. It only has support for deno.land/x right now. I would like this feature to support more registries (like nest.land or jspm.dev). This is currently not practical to do because each registry requires a separate code path in the extension.

As a solution to this I propose a dynamic import IntelliSense that is not specific to any one registry. You should be able to dynamically add more registries without updates to the VS Code Deno extension.

### How does this work for the user?

All registries that want to support this feature must host a `/.well-known/deno-import-intellisense.json` file on their domain (contents of this file are described below). The extension will request this file when it encounters a domain in an import statement for the first time. If the extension finds this file and recognizes the syntax, a popup is shown to the user:

```
Do you want to enable import IntelliSense for https://deno.land? Only do this if you trust https://deno.land.

Learn more

[No] [Yes]
```

If the user enables this, this domain will be added to an array of allowlisted domains in the `deno.import_intellisense` setting in VS Code user settings (override per project possible). If a domain is part of this allowlist, autocomplete for that domain will be enabled (how this works is described below).

### How does this technically work?

When an import is found with a matching registry it is parsed according to the schema defined in `/.well-known/deno-import-intellisense.json`. Depending on where in the URL the cursor is, the correct auto complete should be triggered. The actual IntelliSense logic should be the same as the one we use now. 

### Schema of `/.well-known/deno-import-intellisense.json`

This is approximately what the schema could look like for https://deno.land:

```json
{
  "version": 1,
  "registries": [
    {
      "schema": "/x/:module@:version/:path*",
      "variables": {
        "module": {
          "url": "https://api.deno.land/importintellisense/modules"
        },
        "version": {
          "url": "https://api.deno.land/importintellisense/modules/${module}"
        },
        "path": {
          "url": "https://api.deno.land/importintellisense/modules/${module}/versions/${version}"
        }
      }
    },
    {
      "schema": "/std@:version/:path*",
      "variables": {
        "version": {
          "url": "https://api.deno.land/importintellisense/modules/std"
        },
        "path": {
          "url": "https://api.deno.land/importintellisense/modules/std/versions/${{version}}"
        }
      }
    }
  ]
}
```

So what does this mean?

The file contains a list of registries under this domain, each with a url schema it containing variables. These variables have a specified URL endpoint that is used to request the information for auto complete. These urls can have replacement variables which will be populated with the extracted values from matched the URL schema. There are two types of replacement variables: `${}` which inserts the literal value of the variable and `${{}}` which inserts the URL encoded value of the variable. These URLs are expected to return an array of strings in JSON format when sent a HTTP GET request. They will cached (respecting the `cache-control` headers).

cc @kitsonk @t8 @CGQAQ @callionica @David-Else

lucacasonato:
TODO:
- [x] Prompt user when new origin is seen
  - [x] Only try once a day
- [x] Fetch & parse `/.well-known/deno-import-intellisense.json`
  - [x] Tests
- [x] Autocomplete using info from well-known file
  - [x] Tests
- [x] Implement `deno.import_intellisense_discovery` setting
- [x] Cache `/.well-known/deno-import-intellisense.json` to disk
- [x] Cache API calls to disk
- [x] Write document on what to do as a registry to be auto-discovered

Closes #217


lucacasonato:
This PR is now at a stage where it is usable (still missing tests and caching for remote files).

it_works_i_guess

Now that I know this works I will integrate the required endpoints into the deno.land/x registry so we have a real test target.

client/src/extension.ts
```diff
@@ -311,6 +313,29 @@ export class Extension {
 
           return config;
         });
+
+        client.onRequest(
+          Request.promptEnableImportIntelliSense,
+          async (origin: string) => {
+            const resp = await window.showInformationMessage(
+              `Do you want to enable import IntelliSense for ${origin}? Only do this if you trust ${origin}. [Learn more](https://deno.land).`,
```
bartlomieju, line 321: link to `Learn more` should be fixed?

import_intellisense.md
```diff
@@ -0,0 +1,229 @@
...
+file these are described. For the sake of your time we will not discuss these
+further, as they work the same way as the example above.
+
+### Debugging
+
+You're schema might not work on the first attempt. To find out what is not working
+correctly open the VS Code `OUTPUT` panel and select `Deno Language Server`. This
+is where you can find any schema validation issues your wellknown file might have,
+or other errors that happened during completion. If this does not solve your issue,
+please reach out on Discord at https://discord.gg/deno.
```
bartlomieju, line 229: Nice readme

bartlomieju (approved):

PR #221, opened 2020-09-20
Feat: debug

### LINKED ISSUE #212: Add debug support to vscode_deno
* Launch config snippets ( debug, watch mode )
* F5 to debug current file

I'll implement it later today

CGQAQ:
## Debug 
debug demo

Closes #212


CGQAQ:
image
Don't support vscode stable version due to vscode bug, insiders version works well, in case you encountered this issue

CGQAQ:

lucacasonato:
> Don't support vscode stable version due to vscode bug, insiders version works well, in case you encountered this issue

Which bug is this? Do you have a link?

CGQAQ:
https://github.com/microsoft/vscode/issues/106703

lucacasonato:
Thanks - lets split this into two PRs though. The debugging looks good, but I have some comments for the watching.

CGQAQ:
@lucacasonato Splitted

CGQAQ:
> image
> Don't support vscode stable version due to vscode bug, insiders version works well, in case you encountered this issue

@lucacasonato  This bug is bypassed, now it's working on vscode stable version as well

lucacasonato:
Awesome :+1:

I found one little bug, but this might be VS Code. When you create a launch.json via the debug panel, the snippet pieces are not interpreted correctly - they just show up as a literal.

Peek 2020-09-20 18-04

Also one minor thing. If there is no launch.json, and I go the debug panel, press `Run and Debug`, and select `Deno`, the debug icon in the bottom status bar doesn't show any name. If I debug for example this extension it shows `Client + Server`. Is there any way that can be fixed? It is not that important though :-)

client/src/debug_config_provider.ts
```diff
@@ -0,0 +1,85 @@
...
+      ) {
+        // https://github.com/microsoft/vscode/issues/106703#issuecomment-694595773
+        // Bypass the bug of the vscode 1.49.0
+        debug.startDebugging(workspace, {
+          name: "Deno: Debug",
+          request: "launch",
+          type: "pwa-node",
+          program: "${file}",
+          runtimeExecutable: "deno",
+          runtimeArgs: ["run", "--inspect-brk", "-A"],
```
lucacasonato, line 67: Deno always defaults to no permissions, so we should do the same in the extension. People can set up permissions themselves in a `launch.json`.

```suggestion
          runtimeArgs: ["run", "--inspect-brk"],
```

CGQAQ: this one is not for users to change, it's used when you don't have a launch.json and you press f5 and choose deno

CGQAQ: So the user can't change the permission

lucacasonato: Yeah I think that is fine. If they need permissions they can add a `/.vscode/launch.json` file. I do think it would make sense to automatically set the `--unstable` flag if `deno.unstable` is set to true.

CGQAQ:
> I found one little bug, but this might be VS Code. When you create a launch.json via the debug panel, the snippet pieces are not interpreted correctly - they just show up as a literal.

fixed

CGQAQ:
>  If I debug for example this extension it shows Client + Server

The extension debug itself have launch.json

lucacasonato:
> The extension debug itself have launch.json

So you need a `launch.json` for that label to show up? That is ok.

CGQAQ:
> So you need a launch.json for that label to show up? 

ddddd

client/src/debug_config_provider.ts
```diff
@@ -0,0 +1,99 @@
...
+  window,
+} from "vscode";
+import { ConfigurationField } from "../../core/configuration";
+
+export interface DenoCmd {
+  name: string;
+  entrypoint: string;
+  security_flags: string;
+  port: string;
+}
```
lucacasonato, line 17: This isn't used it seems

lucacasonato (approved):
One final nitpick

client/src/debug_config_provider.ts
CGQAQ: oh! forget to delete

CGQAQ:
I think we should add husky back to format and lint the code before commit @lucacasonato

lucacasonato:
> good night!

Good night!

Issue #200, opened 2020-09-11
"A request has failed. See the output for more information" when format fails, and it always shows output as the cancel `x` button is broken


David-Else:
When using the extension to format TypeScript on save,
```
  "[typescript]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
```
it is very common that there is a problem with your code which means it can't complete. This is an advantage of format on save, it can auto-complete things for you, and it can't always be right. In the Prettier extension, there is a little logo at the screen bottom that indicates it can't format.

With the Deno extension it issues the pop-up warning:

"A request has failed. See the output for more information." 

![1](https://user-images.githubusercontent.com/12832280/92934244-28a15480-f43f-11ea-99c3-4cdaf692f26e.png)

If you press the `x` to cancel, then there is a bug as it shows the output anyway:

![2](https://user-images.githubusercontent.com/12832280/92934261-2ccd7200-f43f-11ea-9fbd-7b76953fb2cd.png)

This completely interrupts the workflow as the user now has to close two windows to continue.

**Versions**

vscode: 1.49.0
deno: 1.3.3
extension: 2.2.1

wsmd:
I'm also having a similar issue here! Not only the output panel appears after clicking X/Cancel (which could be VSCode related bug), but also the output panel pops up automatically every time the formatter fails!

I usually keep the panel focused on the integrated terminal, and having vscode-deno takes the panel focus occasionally on formatting errors can be distracting.

I'm not sure if this is intentional, but it doesn't seems to be any error handling around the document formatting logic, which is causing those error to propagate to all the way to VSCode. 

I captured this behavior here (output appears on click X, and also the entire output panel pops up automatically on save at times)

![2020-10-05 01 18 29](https://user-images.githubusercontent.com/2100222/95043175-0ef3d580-06aa-11eb-8796-451a79e935fd.gif)

> In the Prettier extension, there is a little logo at the screen bottom that indicates it can't format.

I also like the Prettier extension's silent approach to handling formatting errors. When it fails it indicates errors using the status bar item. I went ahead and tried to implement a similar behavior here and I like it much better! To see the error, I can click the statusbar item to open the Deno Language Server output.

![2020-10-05 01 22 52](https://user-images.githubusercontent.com/2100222/95043739-aad21100-06ab-11eb-96c9-d970daafe6b0.gif)

I'm happy to open a PR if this sounds reasonable!

axetroy:
[...]

PR #222, opened 2020-09-20
Feat: deno watch

CGQAQ:
cherry-pick feat_deno-watch out of feat_debug

## Watch
deno watch


wsmd:
I was thinking that it would be nice to the use the status bar item to display error/warn status of the Deno LS (possibly bind the status bar item to the output panel as well), so I'm curious to see how this feature can work with this proposal https://github.com/denoland/vscode_deno/issues/200#issuecomment-703413454

CGQAQ:
> This is a nice addition, but having this feature built into the extension itself has some limitations and needs some thinking through; how do we pass arguments or environment variables? what if the program requires Deno permissions? what about importmaps?

This feature just aim to click and go as playground test environment, if you want to pass arguments, just do what you normally would do

lucacasonato:

PR #229, opened 2020-09-24
typescript-deno-plugin reborn

CGQAQ:
## Currently:

Y0LBpUToqc


CGQAQ:
@lucacasonato I tried a lot, But still cant load/unload d.ts dynamically, so I think we should prompt users every time the config changes to `reload window`

PR #242, opened 2020-10-14
chore: add `deno init` to usage

zhmushan:


lucacasonato:

PR #258, opened 2020-11-08
fix: add step to configure unstable mode

### LINKED ISSUE #257: deno.unstable setting is true by default when using the command palette entry "deno: Init"
When selecting the VS Code command palette entry "deno: Init", the resulting data is added to `.vscode/settings.json`:

```jsonc
{
  "deno.enable": true,
  "deno.lint": false, // depends on selection
  "deno.unstable": true // <-- this line
}
```

This is undesirable because unstable features should be opt-in, NOT the default.

**To Reproduce**

1. <kbd>shift</kbd> + <kbd>cmd</kbd>/<kbd>ctrl</kbd> + <kbd>p</kbd>
2. <kbd>d</kbd> <kbd>e</kbd> <kbd>n</kbd> <kbd>o</kbd> <kbd>i</kbd> <kbd>n</kbd> <kbd>i</kbd> <kbd>t</kbd> <kbd>Enter</kbd>
3. Make a choice about using deno lint
4. (Resulting data is added to file)

**Expected behavior**

`deno.unstable` to be `false`, but, instead it is `true`

**Versions**

vscode: `1.51.0`
deno: `1.5.1`
extension: `2.3.2`

RA80533:
The pull request adds an additional step to the "deno: Init" command to configure unstable mode based on the user's input.

Fixes #257

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -514,8 +514,7 @@ Executable ${this.denoInfo.executablePath}`;
         // if you init project, you enable the plugin. NO NEED CHOOSE
         await config.update("enable", true);
         await config.update("lint", setting.lint);
-        // if lint enabled, enable unstable as well
-        await config.update("unstable", setting.lint);
+        await config.update("unstable", setting.unstable);
         window
           .showInformationMessage(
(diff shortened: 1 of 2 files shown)


lucacasonato:
Thanks for the PR, but we are doing a full extension rewrite so this is not needed anymore.

PR #263, opened 2020-11-27
Version 3 of Extension which uses forthcoming Deno Language Server

kitsonk:
This is still a work in progress, but this branch "works" with builds of Deno that include the language server.

Things that still need to be done on this branch:

- [x] fixup CI, ideally we use the Deno CLI for everything like formatting, linting, etc.  Several of the `package.json` scripts now use this, but we need a version of Deno in the CI enviornment.
- [x] ~use `deno test` for testing.~ skipping for now, nothing to test
- [x] double check the packaging and information.
- [x] update all the documentation and readme with the new version information.


lucacasonato:
@kitsonk I think this is ready to land now. I have some usability improvements I want to make, but I would rather do that in a follow-up PR.

bartlomieju (approved):
LGTM, it's nice to see to many files go 👍

PR #266, opened 2020-12-01
feat: add suppressing quick info and document highlights

kitsonk:
The language server can now provide hover/quick info information and document highlights, so should be suppressed when Deno is enabled on a workspace.


lucacasonato (approved):
LGTM.

PR #270, opened 2020-12-04
feat: suppress references and definitions

kitsonk:
Because these are becoming available in language server.


lucacasonato (approved):
LGTM

PR #285, opened 2020-12-11
fix: include typescript-deno-plugin in vsix

lucacasonato:
Because it is not currently included, the plugin is not installed in the released VS Code extension, which means double diagnostics.


bartlomieju (approved):
LGTM

kitsonk:
🤦

PR #291, opened 2020-12-29
feat: add deno cache command

kitsonk:
Requires denoland/deno#8911


lucacasonato (approved):
LGTM. I will make a PR for `textDocument/rename`, then we can do a 0.0.6 release.

PR #304, opened 2021-01-22
0.0.7

lucacasonato:


kitsonk (changes requested):
Do an npm install so that the package lock gets updated, otherwise there is a delta when you then install it.

kitsonk (approved):
LGTM

PR #312, opened 2021-02-02
feat: add applyCodeActionCommand command

kitsonk:
Plus various other fixes and cleanups.


lucacasonato (approved):
LGTM, but don't quite understand what this does yet. Seems I need to research lsp code actions some more.

kitsonk:
Certain code actions served up by TypeScript provide state modifications, beyond just the text file edits, that need to be applied to TypeScript compiler when the code action is applied in the editor.  In order to make sure those get back to the TypeScript compiler, the client needs to support a command that passes those back to the server so they can be sent to TypeScript.  The whole thing is a rather complex flow, but it mimics what the built in TypeScript extensions does in VSCode.

PR #315, opened 2021-02-04
fix: remove deno/applyCodeActionCommand

kitsonk:
Looks like I was a bit premature.  We don't need `deno/applyCodeActionCommand` because we won't be processing any of the the fixes that include the command.  The only command at the moment is to do an npm package install.  It also works asynchronously inside of tsc, which our current integration doesn't support, and so it was easier to remove the functionality from the Deno lsp that make it work and then do nothing.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -12,10 +12,7 @@ import {
   workspace,
 } from "vscode";
 import { LanguageClient, Location, Position } from "vscode-languageclient";
-import {
-  applyCodeAction as applyCodeActionReq,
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM, but it would be great to have this infra for a module download quick fix in the future.

kitsonk:
> but it would be great to have this infra for a module download quick fix in the future

We already have it... `deno/cache`.  I hopefully will have a good action for that soon.  This was only to send npm package install commands to TypeScript, not useful for us.

PR #316, opened 2021-02-05
feat: add initialize workspace command

kitsonk:
This add the initialize workspace command which was simliar to the init command in v2.

It also does additional prep and changes to make this the released branch instead of canary, and updates to vscode-languageserver 7.0 which moves to LSP 3.16 which is what the Deno CLI LSP server is based on.


lucacasonato (approved):
LGTM!

PR #330, opened 2021-02-18
feat: use preview instead of display for status

kitsonk:
Instead of previewing the Deno LSP status virtual document as a markdown file, it opens it in markdown preview.  This will allow us to provide a more "readable" version of the status page in the future without having to look directly at markdown.

_Extension_Development_Host__-_Preview_status_md_—_vs_test


lucacasonato (approved):
We should do another release of the extension this week.

kitsonk:
> We should do another release of the extension this week.

I was planning on it after 1.7.5 goes out and we convert it to 3.0.

PR #340, opened 2021-02-20
docs: recommend import_map.json instead of import-map.json

### LINKED ISSUE #338: Update import map name to reflect conventional usage
### test

**Describe the bug**

The extension recommends `import-map.json` with a **dash**, in a few places.

It would be preferable to recommend `import_map.json` with an **underscore**, to match the Deno docs - [Import maps](https://deno.land/manual/linking_to_external_code/import_maps).

This is not a big deal, but I am fan of consistency. I'd rather encourage users to use the same convention as in the official Deno docs. And avoid someone overlooking a hyphen and an underscore and scratching their head at results.

**To Reproduce**

This affects someone reading the help text / docs in README.md.

Also, when editing the settings, start editing the import map field and use the auto-fill feature, or wait for the help text to pop up.

> The file path to an import map. This is the equivalent to using `--import-map` on the command line

**Expected behavior**

All messaging in the settings help and  the docs should to `import_map.json` as the first option.

**Screenshots**

Deno docs:

docs

Hover:

hover

Autofill

autofill

Search for affected areas: [q=import-map](https://github.com/denoland/vscode_deno/search?q=import-map)

**Versions**

vscode: not relevant
deno: 1.7 but not relevant
extension: v3.0.1

Liamolucko:
Resolves #338

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,8 @@ extension has the following configuration options:
   [Import maps](https://deno.land/manual/linking_to_external_code/import_maps)
   provide a way to "relocate" modules based on their specifiers. The path can
   either be relative to the workspace, or an absolute path. _string, default
-  `null`, examples: `./import-map.json`, `/path/to/import-map.json`,
-  `C:\path\to\import-map.json`_
+  `null`, examples: `./import_map.json`, `/path/to/import_map.json`,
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

PR #350, opened 2021-03-02
feat: add deno.path setting

kitsonk:
This allows setting a specific path to the `deno` executable for the extension, in either the workspace or globally, instead of being dependent on `deno` being in the environment path only.


lucacasonato (approved):
LGTM. Thanks!

Issue #233, opened 2020-09-24
Debugger not working, Unable to retrieve source content from standard library


David-Else:
I press f5 in `mod.ts` to debug, choose deno, but then a file called `colors.ts` appears with:

`Could not load source '/home/xx/workspace-directory/std@0.70.0/fmt/colors.ts': Unable to retrieve source content.`

The repo is https://github.com/David-Else/eslint-config-octopus and video:
![out](https://user-images.githubusercontent.com/12832280/94148198-38249280-fe6e-11ea-99ea-e9f39f98de82.gif)

**Versions**

vscode: 1.49.1
deno: 1.4.1
extension: 2.3.1

Sharpiro:
I'm getting this error with the sample web server from the manual

https://deno.land/manual/examples/http_server

`Could not load source '/home/user/src/app/server/std@0.71.0/encoding/utf8.ts': Unable to retrieve source content.`

algiuxass:
For me debugger is working but debugger stops at `std/fmt/colors.ts:44` like there's a breakpoint there, but I have no breakpoints enabled and disabled uncaught exceptions breakpoints.

Sometimes debugger shows filenames like `colors.ts>`, `mod.ts>` with `>` at the end. Sometimes it does not.

David-Else:
@algiuxass Have you tried with the `launch.json` from https://deno.land/manual@v1.5.3/getting_started/debugging_your_code#vscode?

mvolkmann:
@David-Else I looked over the notes at the site you suggested, but it didn't help. I can run the debugger. If I step into an import module, I get a popup that says "This is a missing file path referenced by a sourcemap. Would you like to debug the compiled version instead?". If I click "Yes" then I can step through the compiled code. So there are two issues:

1) Why does it ask me that question for every imported source file I step into?
2) Why doesn't it recognize the "//# sourceMappingURL" comment at the bottom of those files and let me step through the TypeScript code of the imported source file?

Here is my `.vscode/launch.json` file inside my Deno project:

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Deno",
      "type": "pwa-node",
      "request": "launch",
      "cwd": "${workspaceFolder}",
      "runtimeExecutable": "deno",
      "runtimeArgs": [
        "run",
        "-A",
        "--inspect-brk",
        "${file}"
      ],
      "attachSimplePort": 9229,
      "outputCapture": "std"
    }
  ]
}
```

[...]

PR #351, opened 2021-03-02
feat: read-add debug support

kitsonk:
This re-adds back support for out of the box debug support, which we had in the 2.X branch.


lucacasonato (approved):
LGTM. PR title is incorrect though.

Soremwar:
It seems to me like the default for this command should be `--inspect` instead of `--inspect-brk`.

Issues like this https://github.com/denoland/vscode_deno/issues/233 show that there is a general missunderstanding about how a debugging session should start. Even though the issue in itself reports another error, users come to think that the debugger is not working because it stops in the first line instead of launching it as deno run would normally do

Issue #361, opened 2021-03-07
Extension prefers resolving to `deno.exe` over resolution of `deno` command on Windows


dsherret:
I use my [binary version manager](https://github.com/bvm/bvm) to manage versions of Deno. On windows, this works by having a `deno.bat` and `deno.ps1` file on the path instead of a `deno.exe`.

The issue at the moment, is that when the extension spawns the command "deno", it's not resolving based on the path and instead selects the ".exe" file found on the path (I think it's ignoring other file types in pathext).

**Reason**

This is fundamentally because of the behaviour of `spawn` in `child_process` that is called in vscode-languageserver-node.

https://github.com/microsoft/vscode-languageserver-node/blob/54b686f0a1817a845f34bda19d5b1651c445f2cf/client/src/node/main.ts#L427

I don't understand or know enough about it, but for example, give the following code:

```js
// test.js
const cp = require("child_process");

const denoCmd = cp.spawn("deno", ["-V"], {
    // shell: true, // works when this line is uncommented
});

denoCmd.stdout.on("data", data => {
    console.log(data.toString());
});
```

...when I have `deno` on the path as 1.8.0, but also have a `deno.exe` deeper down as version 1.5.2, the following occurs:

```bash
deno -V # deno 1.8.0
node test.js # deno 1.5.2
```

However, when I set `shell: true` they will both have the same output.

**Fix**

The fix for this would be for the user to delete `deno.exe` from the path, or to provide `shell: true` as one of the options for `Executable` in the extension:

https://github.com/denoland/vscode_deno/blob/1c9cc2bdef4a5c9cdc3df0c3dd77915b5b4f245b/client/src/extension.ts#L85

I feel like the second would be desirable, but I'm unsure about the side effects. I feel like it wouldn't be a big deal to add this change though... thoughts? I could open a PR.

kitsonk:
What would be the drawback of using the `deno.path` configuration option in this scenario?

dsherret:
@kitsonk The drawback would be having to configure something that should be automatic. Also, it’s not immediately clear to users that it’s potentially resolving to a different version of Deno than what’s found on the path.

By the way, a third option for a fix is to resolve the correct path based on pathext and the current directories found on the path. I don’t know why spawn doesn’t respect pathext, but seems like that’s why this package exists: https://www.npmjs.com/package/cross-spawn

kitsonk:
Ok.  If you don't mind opening the PR.

dsherret:
[...]

PR #367, opened 2021-03-12
feat: Manual resolution of `deno` command on Windows

dsherret:
See issue #361 for more details.

This ensures the `deno` command automatically resolves the same way it resolves on the command line.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -9,6 +9,9 @@ import {
 import type { Settings } from "./interfaces";
 import { DenoTextDocumentContentProvider, SCHEME } from "./content_provider";
 import { DenoDebugConfigurationProvider } from "./debug_config_provider";
+import * as fs from "fs";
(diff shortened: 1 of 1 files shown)


client/src/extension.ts
```diff
@@ -78,7 +81,8 @@ export async function activate(
   context: vscode.ExtensionContext,
 ): Promise<void> {
   const command =
-    vscode.workspace.getConfiguration("deno").get<string>("path") || "deno";
+    vscode.workspace.getConfiguration("deno").get<string>("path") ||
+    await getDefaultDenoCommand();
```
dsherret, line 85: Side note: I think leading logical operators help break this up better. This almost looks like a separate statement.

```ts
|| await getDefaultDenoCommand()
```

kitsonk (approved):
LGTM

Thanks.  This should help win folks a lot.

PR #397, opened 2021-04-18
feat: add environment variables support on deno.path

### LINKED ISSUE #377: deno.path doesn't resolve environment variables
The setting "deno.path" doesn't seem to resolve environment variables.

In my settings.json I want to use an environment variable to resolve to deno.

Basically what I would like it to work with:
```"deno.path": "${env:USERPROFILE}\\Applications\\deno\\deno.exe"```
or
```"deno.path": "%USERPROFILE%\\Applications\\deno\\deno.exe"```

But it does work with an absolute path:
```"deno.path": "C:\\Users\\<username>\\Applications\\deno\\deno.exe"```

I tried this with the global settings.json and a workspace file.

**Expected behavior**

The environment variable should be resolved and enable the extension to find deno.

**Screenshots**

image

image

**Versions**

vscode: 1.54.3 (user setup)
deno: 1.8.2
extension: 3.2.0
os: Win10 build 19042

ablause:
Adds the possibility to use environment variables on the Deno path in settings.

Resolves #377

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -227,6 +227,8 @@ async function getCommand() {
   const defaultCommand = await getDefaultDenoCommand();
   if (!command || !workspaceFolders) {
     command = command ?? defaultCommand;
+  } else if ((/[%|$]/g).test(command)) {
+    command = resolveEnvPath(command);
   } else if (!path.isAbsolute(command)) {
     // if sent a relative path, iterate over workspace folders to try and resolve.
     const list = [];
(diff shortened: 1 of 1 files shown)


kitsonk:
Thanks for this, but I think the challenge is that it sort of provides some "bespoke" variable resolution.  I would really prefer to be able to leverage something directly from the vscode platform versus trying to build something ourselves.  There is this really old issue to add it to vscode which would be consistent with other parts of vscode, like the `launch.json`: https://github.com/microsoft/vscode/issues/2809

ry:
Closing because stale.

PR #402, opened 2021-04-25
feat: Add test code lens

keroxp:
- Implemented CodeLens features for deno tests based on https://github.com/firsttris/vscode-jest-runner
- Two features included in this PR.
- Using typescript compiler api for extracting test expression (top-level only)

**Run/Debug button for each test**
image

**Run/Debug file in Context Menu**
image

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -149,7 +149,7 @@ export async function activate(
   );
   const testCodelensEnabled = settings.codeLens?.tests ?? false;
   if (testCodelensEnabled) {
-    const codeLensProvider = new DenoTestCodeLensProvider();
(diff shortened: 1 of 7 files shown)


kitsonk:
Thanks for the contribution, but I would prefer to add this code lens directly in the LSP versus as something that would only work with the VSCode extension.  Adding it only to the vscode extension really reduces the re-usability.

keroxp:
@kitsonk Sure, I'm going to make PR for deno lsp.

kitsonk:
@keroxp thank you very much for trying to contribute this feature.  It required some fundamental shifts in both the lsp and the client as part of the overall extension/lsp roadmap and so I have raised the client part of it under #436.

Closing in favour of #436.

PR #411, opened 2021-05-04
feat: handle per resource configuration

kitsonk:
This PR modifies the client and the built in TypeScript language service plugin to allow "per resource" configuration.  This is required to support VSCode Workspace Folders, which allow different settings per folder in a workspace.

At the moment the only setting supported per resource is `deno.enable`, and it is the only one that effects the behaviour of the extensions language service plugin.  Additional "per resource" scoped features can be enabled as they are enabled within the Deno language server.


lucacasonato:
Tested locally with a Node and a Deno project in workspace folders, and it works great :-)

lucacasonato (approved):

PR #425, opened 2021-06-01
Fix truthy config

Buraburaite:
This PR is for this issue: https://github.com/denoland/deno/issues/10808

Diagnosis
---
If in Settings, someone puts an empty string into config, the `??` will evaluate that to `true` rather than short-circuit to `value.defaultValue` (which I believe is `null`, and comes from `package.json`).

Replication
---
This is my first time editing a vscode extension, so the way that I verified that the fix works is by going to `${HOME}/.vscode-oss/extensions/denoland.vscode-deno-3.4.0/client/out/extension.js` and making the change. Now, even if config is an empty string (with the orange indicator line showing), it fixes the issue with language server.

I don't know if this would be your preferred solution, or if you'd prefer an exception specifically for the config variable, or if I should instead patch lsp. Let me know and I can update it.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -110,7 +110,6 @@ function configToResourceSettings(
       value.globalValue ||
       value.defaultValue;
   }
-  resourceSettings['config'] = resourceSettings['config'] || false;
   return resourceSettings;
 }
 
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -80,17 +80,17 @@ function configToWorkspaceSettings(
   for (const key of workspaceSettingsKeys) {
     const value = config.inspect(key);
     assert(value);
-    workspaceSettings[key] = value.workspaceLanguageValue ??
-      value.workspaceValue ??
-      value.globalValue ??
+    workspaceSettings[key] = value.workspaceLanguageValue ||
+      value.workspaceValue ||
+      value.globalValue ||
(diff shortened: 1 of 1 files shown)


satyarohith:
Hi @Buraburaite. :) We should ideally fix the bug in `deno lsp`. Somewhere around here: https://github.com/denoland/deno/blob/f8913680573505e45f0c094af1421f7a4abb4fb2/cli/lsp/language_server.rs#L425

kitsonk:
Correct, because it being `null` or `undefined` or `""` or any non-truthy value should not cause the value to be loaded. This needs to be fixed in the Deno language server.  In addition the nullish-coalescing is very intentional to ensure that if there is a value set, it is properly reflected, even a "falsey" one.

Thanks, but closing without merging.

Buraburaite:
Hi @satyarohith,

I would actually love to do this! But I don't know Rust. I'm watching some videos and will submit a PR for lsp once I'm sure my code is doing what I think it's doing.

That being said, if the solution isn't something someone with more fluency would write, hopefully you or someone else could take it from there and I could learn from the example.

Thanks

satyarohith:
> I would actually love to do this! But I don't know Rust. I'm watching some videos and will submit a PR for lsp once I'm sure my code is doing what I think it's doing.

Please take your time. I think it should be straightforward to fix. :)

PR #429, opened 2021-06-01
fix: bump semver of extension

kitsonk:
Sets the semver to `>=1.10.3` which includes the ability to format `.md`, where in earlier versions will cause strange diagnostics.


kitsonk:
I propose we we cherry-pick this into a `3.5.1` release (because there are other features now in main that aren't in the released version of Deno).

lucacasonato (approved):

PR #436, opened 2021-06-07
feat: add support for tasks and test code lens

kitsonk:
This adds a task provider for Deno CLI tasks as well as supports running individual tests based on a test code lens provided by the language server.  This requires https://github.com/denoland/deno/pull/10874 to be useful.


lucacasonato:
Implementation looks good, but running into some issues while playing around with it:

Create a new file:

```ts
const foo: number = "";
```

Add a task to run it:

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "type": "deno",
      "command": "run",
      "problemMatcher": [
        "$deno"
      ],
      "args": ["main.ts"],
      "label": "deno: run"
    }
  ]
}
```

Before running, see that a diagnostic for the TSC error shows up on the problem pane.

Now run the task. The "problem" diagnostic from deno-tsc disappears from the problem pane. There are now no problems anymore.

kitsonk:
Ok, it seems that it is related to the "ownership" of problems.  If you comment out the problem matcher the clearing goes away.  I am going to change the "problem" owner for the CLI, so that the LSP and the CLI don't overlap.

lucacasonato (approved):

PR #446, opened 2021-06-08
feat: add support for import map in test code lens

cvng:


kitsonk (approved):
LGTM. Great idea. I guess we should add it for the config option too.

PR #452, opened 2021-06-18
fix: setting then clearing "deno.path" config should allow language server to start

### LINKED ISSUE #451: Setting then clearing "deno.path" config causes lsp to not start
**Describe the bug**

Setting then clearing the "deno.path" config setting via the GUI causes the deno extension to fail to start.

**To Reproduce**

1. Open up the GUI settings editor and add a user setting for "Deno: Path"
2. Save. Then clear that value.

**Expected behavior**

**Output**

```
Activating extension 'denoland.vscode-deno' failed: Unsupported server configuration {
    "run": {
        "command": "",
        "args": [
            "lsp"
        ],
        "options": {
            "env": {
               // omitted
            }
        }
    },
    "debug": {
        "command": "",
        "args": [
            "lsp"
        ],
        "options": {
            "env": {
               // omitted
            }
        }
    }
}.
```

```
Couldn't start client Deno Language Server
```

**Versions**

vscode: 1.57.1 deno: 1.11.0 extension: 3.6.1

dsherret:
Closes #451.

The reason this occurs is because the GUI sets the value in the settings json to `"deno.path": ""`

### DIFF
--- a/client/src/util.ts
+++ b/client/src/util.ts
@@ -18,7 +18,7 @@ export async function getDenoCommand(): Promise<string> {
   if (memoizedCommand !== undefined) {
     return memoizedCommand;
   }
-  let command = vscode.workspace.getConfiguration("deno").get<string>("path");
+  let command = getWorkspaceConfigDenoExePath();
   const workspaceFolders = vscode.workspace.workspaceFolders;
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM

Issue #374, opened 2021-03-20
lsp sometimes turns into zombie process


tyler338:
**To Reproduce**
1. open vscode
2. play around in editor for like a minute
3. close vscode with cmd+q
4. lsp turns into zombie process

*note: doesn't always happen consistently, so it might require multiple attempts or different steps*

**Expected behavior**
lsp exits with vscode

**Screenshots**
![](https://cdn.discordapp.com/attachments/684898665151594506/822791887357345822/unknown.png)
**Versions**

vscode: 1.54.3 (Universal)
deno: 1.8.1+0e70d9e
extension: 3.2.0

qwerasd205:
I have also experienced this issue (macOS)
It seems to occur most often (or perhaps most noticably) for me when I have large/complex javascript files open (For example, large obfuscated samples) which leaves the process using 100% of one of my CPU cores after quitting VS Code.

imcotton:
Same here on **macOS**,  dangling `deno lsp` process each time open/close **vscode** while extension enabled within **Deno** project.

Right now I'm using following as temporary cleanup solution:

    pkill -f 'deno lsp'

kidonng:
Just noticed this issue when browsing Activity Monitor. Literally freed 1GB worth of RAM after `killall deno`.

![image](https://user-images.githubusercontent.com/44045911/119254091-fb6c8f00-bbe6-11eb-9e40-a430244c5316.png)

There should be some kind of notice of this issue, or it is too subtle to to notice.

KnorpelSenf:
Happens to me regularly when just _reloading_ VSCode (`Developer: Reload Window`). The extension even starts loads of zombie processes when I explicitly set `deno.enable: false` (before starting VSCode) in `.vscode/settings.json`. It's been this way for half a year or so, for multiple versions of both `deno` and this extension. Current versions are:

```bash
$ uname -a
Linux steffens-pc 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux
$ deno --version 
deno 1.10.3 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.27
typescript 4.2.2
$ code --version
1.56.2
054a9295330880ed74ceaedda236253b4f39a335
x64
```

jeiea:
Same on windows.
```
> code --version
1.56.2
054a9295330880ed74ceaedda236253b4f39a335
x64
> deno --version
[...]

PR #453, opened 2021-06-21
fix: use parent-pid flag to prevent orphaned processes

dsherret:
We need to run `deno -V` to figure out the version because we can't call `deno lsp --parent-pid <pid>` on an old version of Deno.

Had to do a bit of a refactor to implement this.

Relates to #374

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -17,15 +17,21 @@ import {
 import * as tasks from "./tasks";
 import type { DenoExtensionContext } from "./types";
 import { WelcomePanel } from "./welcome";
-import { assert } from "./util";
+import { assert, getDenoCommandAndVersion } from "./util";
+import { registryState } from "./lsp_extensions";
+import { createRegistryStateHandler } from "./notification_handlers";
 
+import * as semver from "semver";
 import * as vscode from "vscode";
-import { LanguageClient } from "vscode-languageclient/node";
+import { LanguageClient, ServerOptions } from "vscode-languageclient/node";
 import type {
   DocumentUri,
   Location,
   Position,
 } from "vscode-languageclient/node";
+
+/** The minimum version of Deno that this extension is designed to support. */
+const SERVER_SEMVER = ">=1.11.2";
 
 // deno-lint-ignore no-explicit-any
 export type Callback = (...args: any[]) => unknown;
@@ -42,14 +48,15 @@ export function cache(
 ): Callback {
   return (uris: DocumentUri[] = []) => {
     const activeEditor = vscode.window.activeTextEditor;
-    if (!activeEditor) {
+    const client = extensionContext.client;
+    if (!activeEditor || !client) {
       return;
     }
     return vscode.window.withProgress({
       location: vscode.ProgressLocation.Window,
       title: "caching",
     }, () => {
-      return extensionContext.client.sendRequest(
+      return client.sendRequest(
         cacheReq,
         {
           referrer: { uri: activeEditor.document.uri.toString() },
@@ -86,7 +93,7 @@ export function reloadImportRegistries(
   _context: vscode.ExtensionContext,
   { client }: DenoExtensionContext,
 ): Callback {
-  return () => client.sendRequest(reloadImportRegistriesReq);
+  return () => client?.sendRequest(reloadImportRegistriesReq);
 }
 
 /** Start (or restart) the Deno Language Server */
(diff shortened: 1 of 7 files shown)


client/src/commands.ts
```diff
@@ -17,16 +17,22 @@ import {
...
-import { LanguageClient } from "vscode-languageclient/node";
+import { LanguageClient, ServerOptions } from "vscode-languageclient/node";
 import type {
   DocumentUri,
   Location,
   Position,
 } from "vscode-languageclient/node";
 
+/** The minimum version of Deno that this extension is designed to support. */
+const SERVER_SEMVER = ">=1.11.2";
```
dsherret, line 34: I moved a lot of code in here from extension.ts so that it can better handle error paths and do an actual reset when running the `Restart Language Service`.

client/src/commands.ts
```diff
@@ -42,14 +48,15 @@ export function cache(
 ): Callback {
   return (uris: DocumentUri[] = []) => {
     const activeEditor = vscode.window.activeTextEditor;
-    if (!activeEditor) {
+    const client = extensionContext.client;
+    if (!activeEditor || !client) {
```
dsherret, line 52: I have it returning silently when the language service doesn't exist here and in other places. Perhaps it should throw an error instead?

client/src/extension.ts
```diff
@@ -266,32 +245,16 @@ export async function activate(
...
   extensionContext.tsApi = await getTsApi();
-
-  await commands.startLanguageServer(context, extensionContext)();
-
-  context.subscriptions.push(
-    extensionContext.client.onNotification(
-      registryState,
-      createRegistryStateHandler(),
-    ),
-  );
```
dsherret, line 277: Moved into the start language service client command.

client/src/types.d.ts
```diff
@@ -57,11 +56,10 @@ export interface DocumentSettings {
 }
 
 export interface DenoExtensionContext {
-  client: LanguageClient;
+  client: LanguageClient | undefined;
```
dsherret, line 59: When the language server doesn't start then this will be undefined.

client/src/util.ts
```diff
@@ -12,12 +15,34 @@ export function assert(cond: unknown, msg = "Assertion failed."): asserts cond {
...
+  } catch (err) {
+    console.error(`Error getting deno version: ${err}`);
   }
+  return {
+    command,
+    version,
+  };
+}
+
+export async function getDenoCommand(): Promise<string> {
```
dsherret, line 45: I don't think we should memoize this as it might prevent us from getting the latest information (I removed the memoizing). This is only used when starting the language service anyway.

dsherret:
It would be good to manually test this out more before we release as there are a decent amount of changes here.

kitsonk:
Seeing it like this begs the question of why we can't set the ppid during the initialization flow with the language server and why it has to be passed on the CLI?  Having it part of the initialization flow would mean other clients could implement it without much of a problem and non-supportive version of Deno would simply ignore it.  Having to do a `deno -v` seems like a bodge.

dsherret:
@kitsonk I'm not too familiar with the LSP, but that does sound a lot better. How would it work? Is there some code that already does something similar that I'm missing?

kitsonk:
@dsherret a client sends an initialisation message with [InitializeParams](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#initializeParams) with its capabilities and the server provides back its [capabilities](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#serverCapabilities) and there is syntax for supporting custom capabilities.

> That said, it might still be useful to do a `deno -V` beforehand in order to give better error messages. The vs code language server client didn't seem to be doing a good job of that (other than for saying a version was too old).

I don't see `deno -v` solving that.  If the error messages can be improved, it can be done so without invoking `deno -V`.

> Edit: I was just thinking a bit more about it right now and the one benefit a CLI flag brings is that it's a bit more resilient to failure. We can be sure when starting the process that it received the pid.

I'm not sure I agree with that.  We can investigate what happens if the initialisation flow the LSP doesn't setup properly, but I can't see it being passed on the CLI as any more robust.

It certainly would be much better to embrace the capability model of the language server protocol than version sniffing which is fragile, and even if we are version sniffing, that is part of the protocol.

Specifically there is the `"experimental"` server capabilities where this specific capability could be published as being supported:

https://github.com/denoland/deno/blob/388274e02ea7e054f0bbbe4f863dc32447c5971d/cli/lsp/capabilities.rs#L144

And then the client can send a custom `deno/ppid` message, similar to the way we handle other [custom messages](https://github.com/denoland/deno/blob/main/cli/lsp/lsp_custom.rs).

dsherret:
Cool. Thanks! I’ll look into it tomorrow.

dsherret:
@kitsonk actually, given the nature of this bug, do you think we should land this and have this flag passed specifically only for 1.11.2? Then when 1.12 comes out it can have the message passed through the lsp and we can remove the version sniffing? What do you think?

dsherret:
Did some more investigation and the LSP already sends the parent process id in the initialization request!

https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#initializeParams

I'll open a PR in Deno's repo soon.

PR #454, opened 2021-06-22
fix: better handling when language server fails to start

dsherret:
This is essentially #453, but without the `parent-pid` changes.

These changes also...

1. Allow users to update deno then just do the "Restart Language Service" command.
2. Allow users to update "deno.path" and have the changes take effect immediately.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -20,6 +20,7 @@ import { WelcomePanel } from "./welcome";
 import { assert, getDenoCommand } from "./util";
 import { registryState } from "./lsp_extensions";
 import { createRegistryStateHandler } from "./notification_handlers";
+import { configurePlugin } from "./ts_api";
 
 import * as semver from "semver";
 import * as vscode from "vscode";
@@ -29,7 +30,6 @@ import type {
   Location,
   Position,
 } from "vscode-languageclient/node";
-import { configurePlugin } from "./ts_api";
 
 /** The minimum version of Deno that this extension is designed to support. */
 const SERVER_SEMVER = ">=1.10.3";
@@ -106,10 +106,11 @@ export function startLanguageServer(
     // Stop the existing language server and reset the state
     const { statusBarItem } = extensionContext;
     if (extensionContext.client) {
-      await extensionContext.client.stop();
+      const client = extensionContext.client;
       extensionContext.client = undefined;
       statusBarItem.hide();
       vscode.commands.executeCommand("setContext", ENABLEMENT_FLAG, false);
+      await client.stop();
     }
 
     // Start a new language server
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -8,6 +8,7 @@ import {
   EXTENSION_NS,
   LANGUAGE_CLIENT_ID,
   LANGUAGE_CLIENT_NAME,
+  SERVER_SEMVER,
 } from "./constants";
 import { pickInitWorkspace } from "./initialize_project";
 import {
@@ -29,9 +30,6 @@ import type {
   Location,
   Position,
 } from "vscode-languageclient/node";
-
-/** The minimum version of Deno that this extension is designed to support. */
-const SERVER_SEMVER = ">=1.10.3";
 
 // deno-lint-ignore no-explicit-any
 export type Callback = (...args: any[]) => unknown;
@@ -91,9 +89,9 @@ export function initializeWorkspace(
 
 export function reloadImportRegistries(
   _context: vscode.ExtensionContext,
-  { client }: DenoExtensionContext,
+  extensionContext: DenoExtensionContext,
 ): Callback {
-  return () => client?.sendRequest(reloadImportRegistriesReq);
+  return () => extensionContext.client?.sendRequest(reloadImportRegistriesReq);
 }
 
 /** Start (or restart) the Deno Language Server */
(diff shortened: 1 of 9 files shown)


client/src/commands.ts
```diff
@@ -17,16 +17,23 @@ import {
...
-import { LanguageClient } from "vscode-languageclient/node";
+import { LanguageClient, ServerOptions } from "vscode-languageclient/node";
 import type {
   DocumentUri,
   Location,
   Position,
 } from "vscode-languageclient/node";
 
+/** The minimum version of Deno that this extension is designed to support. */
+const SERVER_SEMVER = ">=1.10.3";
```
dsherret, line 34: I moved a lot of code in here from extension.ts so that it can better handle error paths and do an actual reset when running the `Restart Language Service`.

client/src/extension.ts
```diff
@@ -35,22 +20,6 @@ const LANGUAGES = [
...
-  const extension: vscode.Extension<TsLanguageFeatures> | undefined = vscode
-    .extensions.getExtension(TS_LANGUAGE_FEATURES_EXTENSION);
-  const errorMessage =
-    "The Deno extension cannot load the built in TypeScript Language Features. Please try restarting Visual Studio Code.";
-  assert(extension, errorMessage);
-  const languageFeatures = await extension.activate();
-  const api = languageFeatures.getAPI(0);
-  assert(api, errorMessage);
-  return api;
-}
```
dsherret, line 52: Extracted out to its own file and burried some complexity—single `refresh()` method now.

client/src/extension.ts
```diff
@@ -144,7 +106,12 @@ function handleConfigurationChange(event: vscode.ConfigurationChangeEvent) {
         ),
       };
     }
-    configurePlugin();
+    configurePlugin(extensionContext);
+
+    // restart when "deno.path" changes
+    if (event.affectsConfiguration("deno.path")) {
+      vscode.commands.executeCommand("deno.restart");
+    }
```
dsherret, line 114: Restarts the language server when someone changes `"deno.path"`. This is a bit more convenient for users.

Side note: These command names and setting names should probably be moved to constants...

client/src/util.ts
```diff
@@ -27,7 +25,7 @@ export async function getDenoCommand(): Promise<string> {
     // if sent a relative path, iterate over workspace folders to try and resolve.
     const list = [];
     for (const workspace of workspaceFolders) {
-      const dir = path.resolve(workspace.uri.path, command);
+      const dir = path.resolve(workspace.uri.fsPath, command);
```
dsherret, line 28: Bug on windows. Should be `fsPath` in order to remove the leading slash (ex. shoudl be `v:/folder` and not `/v:/folder`).

client/src/types.d.ts
```diff
@@ -57,11 +56,10 @@ export interface DocumentSettings {
 }
 
 export interface DenoExtensionContext {
-  client: LanguageClient;
+  client: LanguageClient | undefined;
```
dsherret, line 64: When the language server doesn't start then this will be undefined.

client/src/util.ts
```diff
@@ -12,12 +15,7 @@ export function assert(cond: unknown, msg = "Assertion failed."): asserts cond {
   }
 }
 
-let memoizedCommand: string | undefined;
```
dsherret, line 15: I don't think we should memoize this as it might prevent us from getting the latest information (ex. `"deno.path"` might have changed). This is only used when starting the language service anyway.

client/src/commands.ts
kitsonk: Maybe move it to: https://github.com/denoland/vscode_deno/blob/main/client/src/constants.ts as it probably should have always been there?

client/src/commands.ts
```diff
@@ -86,7 +93,7 @@ export function reloadImportRegistries(
   _context: vscode.ExtensionContext,
   { client }: DenoExtensionContext,
 ): Callback {
-  return () => client.sendRequest(reloadImportRegistriesReq);
+  return () => client?.sendRequest(reloadImportRegistriesReq);
```
kitsonk, line 96: Feels dodgy, why does TS think this is possibly undefined now?  Feels like we are missing something here.

client/src/extension.ts
kitsonk: 👍

client/src/types.d.ts
kitsonk: Ok, now I see it undefined possibly here...

client/src/commands.ts
kitsonk: Ok, saw below the change in the types, it makes sense here, though are we sure we want to bind early like this?  Don't we want `extensionContext.client?.sendRequest(reloadImportRegistriesReq)` so we always use the currently assigned client?

kitsonk (commented):
looks good, a couple of thoughts

client/src/commands.ts
dsherret: Good catch!

client/src/types.d.ts
dsherret: Sorry for the confusion, I renamed the title to something more clear (and correct).

kitsonk (approved):
LGTM

PR #474, opened 2021-07-23
fix: properly handle plugin configuration at startup

### LINKED ISSUE #473: When activating on a non-supported language document, the typescript-deno-plugin does not init properly
If for example you re-open a workspace and the last file you had open was a JSON file, like an import map, the Deno extension will activate, but then if you change to a tab that is a TypeScript file, the diagnostics from the built in language server are not "muted".

I am not sure exactly what is going on, and I will investigate further, but I think this leads to the reports we have of people complaining about the built in language service not being muted in certain scenarios.

kitsonk:
Fixes #473

### DIFF
--- a/client/src/ts_api.ts
+++ b/client/src/ts_api.ts
@@ -20,22 +20,29 @@ interface TsLanguageFeaturesApiV0 {
   ): void;
 }
 
-export async function getTsApi(
+export function getTsApi(
   getPluginSettings: () => PluginSettings,
-): Promise<TsApi> {
-  const extension: vscode.Extension<TsLanguageFeatures> | undefined = vscode
-    .extensions.getExtension(TS_LANGUAGE_FEATURES_EXTENSION);
-  const errorMessage =
-    "The Deno extension cannot load the built in TypeScript Language Features. Please try restarting Visual Studio Code.";
-  assert(extension, errorMessage);
-  const languageFeatures = await extension.activate();
-  const api = languageFeatures.getAPI(0);
-  assert(api, errorMessage);
+): TsApi {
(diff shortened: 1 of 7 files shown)


client/src/ts_api.ts
```diff
@@ -20,22 +20,29 @@ interface TsLanguageFeaturesApiV0 {
   ): void;
 }
 
-export async function getTsApi(
+export function getTsApi(
```
dsherret, line 23: Does the code that calls this need to be updated to not `await`?

client/src/ts_api.ts
```diff
@@ -20,22 +20,29 @@ interface TsLanguageFeaturesApiV0 {
...
+      .extensions.getExtension(TS_LANGUAGE_FEATURES_EXTENSION);
+    const errorMessage =
+      "The Deno extension cannot load the built in TypeScript Language Features. Please try restarting Visual Studio Code.";
+    assert(extension, errorMessage);
+    const languageFeatures = await extension.activate();
+    api = languageFeatures.getAPI(0);
+    assert(api, errorMessage);
+    const pluginSettings = getPluginSettings();
+    api.configurePlugin(EXTENSION_TS_PLUGIN, pluginSettings);
+  })();
```
dsherret, line 38: What happens if code in here throws for some reason? Does it get surfaced somehow?

dsherret (approved):
LGTM, but a few small things. Sorry I missed this yesterday.

client/src/ts_api.ts
kitsonk: I've added an error message that gets displayed to the user.

Issue #459, opened 2021-06-26
Deno appears in the status bar regardless of whether workspace has "deno.enable": true


dsherret:
I don't see any code that prevents this as right now the commands are registered and then it runs the code to start the language server.

We should register the commands, but only start the language server when the extension is enabled. If someone runs the "Deno: Restart Language Server" command and `"deno.enable"` is not true, then we should tell them to run "Deno: Initialize Workspace".

I checked my global configuration and `"deno.enable"` was not `true`.

![image](https://user-images.githubusercontent.com/1609021/123516696-718f6600-d66b-11eb-97cf-2dc015283893.png)

lucacasonato:
@dsherret The reason we do this is to make the formatteer work even if Deno is not enabled in a workspace (allow the builtin "Deno Formatter" to work in a Node project).

dsherret:
@lucacasonato oh, interesting. I don't have it configured to be used as a formatter though. Perhaps it should only start the language service if it is? Also, perhaps it should be lazily initialized for when a format occurs or instead of starting a background process for formatting it could use stdin and the `deno fmt` command when `"deno:enabled"` is not `true`?

lucacasonato:
@dsherret The extension (and with it the LSP) should only start if you do one of the following right now:

- open a JS, TS, JSON, or MD file
- run a `Deno:` command from the command palette
- start the debugger with Deno configured

I guess what you want to do is only start the LSP in the extension if `deno.enable` is in any configs / commands are triggered, and alternatively fall back to formatting via stdin? I would rather go and profile the LSP and make it more lightwheight for the scenario where `deno.enable` is not set (for example don't start TSC).

dsherret:
I just find it weird that I'm working on something unrelated to Deno, don't have formatting enabled to use Deno, and don't have `"deno.enable": true` in the repo and yet I see it running in the bottom right corner. I'd rather it not start at all in this scenario. Thinking about it more, I guess yeah if someone is using it for formatting then it's annoying to have two ways of doing something. I think in that case to simplify things then we should start it, but not in my case here where it has no use.

kitsonk:
Just like the built in TypeScript plugin runs even if Deno is enabled, there is also state that needs to be tracked so that if someone does enable it, it doesn't require a restart. It gets a lot more complicated with workspace folders where Deno can be disabled and enabled.

I think the only actionable aspect would be to not display the status info on non-Deno enabled projects or on projects where at least one workspace folder is enabled.

[...]

PR #485, opened 2021-08-05
feat: hide the status bar unless `deno.enable` is `true`

dsherret:
This is based on discussion in #459 and we can probably close that one after this.

### DIFF
--- a/client/src/status_bar.ts
+++ b/client/src/status_bar.ts
@@ -1,0 +1,38 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { DenoExtensionContext } from "./types";
+
+import * as vscode from "vscode";
+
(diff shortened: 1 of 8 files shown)


client/src/status_bar.ts
```diff
@@ -0,0 +1,38 @@
...
+      vscode.StatusBarAlignment.Right,
+      0,
+    );
+  }
+
+  dispose() {
+    this.#inner.dispose();
+  }
+
+  refresh(extensionContext: DenoExtensionContext) {
```
dsherret, line 21: I kind of prefer a single state tree then parts of the application listen for state changes and update accordingly. That would be a big refactor though and not sure if desired.

kitsonk (approved):
LGTM and I like how this does some restructuring to improve things.

PR #502, opened 2021-08-24
fix: properly parse config for test code lens

### LINKED ISSUE #494: Cannot read property 'toString' of null
**Describe the bug**
When clicking the `Run Test` code lens I get `Cannot read property 'toString' of null`

**To Reproduce**

1. Create a deno project with linter and unstable features
2. Create a `mod.test.ts` file
3. Fill it with this code

``` typescript
Deno.test({
  name: 'a test case',
  fn() {
    console.log('Foo!');
  },
});
```
4. You will see the `Run Test` above the pasted code.
5. Click it!

**Expected behavior**

The test should run.

**Screenshots**

image

**Versions**

vscode: 1.59.0 deno: 1.13.0 extension: 3.8.0

kitsonk:
Fixes #494

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -229,12 +229,15 @@ export function test(
     if (config.get("unstable")) {
       testArgs.push("--unstable");
     }
-    if (config.has("importMap")) {
+    const importMap: string | undefined | null = config.get("importMap");
+    if (importMap != null) {
       testArgs.push("--import-map", String(config.get("importMap")));
     }
-    const env = config.has("cache")
-      ? { "DENO_DIR": config.get("cache") } as Record<string, string>
-      : undefined;
+    const env = {} as Record<string, string>;
+    const cacheDir: string | undefined | null = config.get("cache");
+    if (cacheDir != null) {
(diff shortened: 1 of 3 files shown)


client/src/commands.ts
```diff
@@ -229,12 +229,15 @@ export function test(
     if (config.get("unstable")) {
       testArgs.push("--unstable");
     }
-    if (config.has("importMap")) {
+    const importMap: string | undefined | null = config.get("importMap");
+    if (importMap != null) {
```
dsherret, line 233: Perhaps check `importMap.trim().length > 0` and the same below?

dsherret (approved):
I think we should maybe consider wrapping `Config` and creating our own interface that's less error prone and gives the user better messages when they do something wrong (ex. when they don't provide a string). There was a similar issue in `getWorkspaceConfigDenoExePath()`.

kitsonk:
@dsherret maybe, it seems we just need to be more careful, it is infrequent we have to deal with it in the client. I mean we don't want to create some enterprise software with loads of arbitrary abstractions 😉 .  But I wouldn't be opposed to it.

PR #503, opened 2021-08-24
chore: bump minimum version of Deno CLI

kitsonk:
This updates to the min version to >= 1.13.0.  While this isn't technically true, v1.13 and later contains several stability and diagnostic fixes which we want to _nag_ people into having when using the LSP.


dsherret (approved):
LGTM and agree.

PR #513, opened 2021-08-30
fix: debug config provider

kitsonk:
This deals with two issues that were causing challenges when configuring debug:

- There was a static config in the `package.json` which then causes the debug provider to not actually work.
- The debug provider did not take into account the context, meaning that if you were using an import map, a cache path, or a config file, none of those would be configured to automatically occur when starting debug or generating a `launch.json`.

In addition, the naming of the launch tasks have been adjusted to be better aligned with `Chrome` and `Node.js` debug configurations provided by vscode.


bartlomieju (approved):
LGTM

dsherret (approved):
LGTM too! Nice change.

PR #514, opened 2021-08-31
fix: better suppression of built in diagnostics

kitsonk:
Trying to fix some minor niggles when supressing diagnostics when Deno is enabled.


dsherret (approved):
LGTM

PR #521, opened 2021-09-14
feat: support deno.json(c) files

### LINKED ISSUE #520: Add support for `deno.json` and `deno.jsonc`
We need to add support for `deno.json` and `deno.jsonc`.

Once we have a schema (see denoland/deno#11885) we should be able to integrate the schema into the plugin.  We also need to update information about the `config` to reflect that it isn't specifically a `tsconfig.json` file.

kitsonk:
Closes #520


bartlomieju (approved):
LGTM

PR #551, opened 2021-09-22
feat: configurable import maps for testing code lens

### LINKED ISSUE #550: Configurable import map for testing code lens
I'm unit testing a library that depends on another library that makes DOM calls. To avoid using flaky DOM implementations, I'm using a separate test-only import map that loads a shimmed version of the said third-party library. That is, I have a general import map with a link to the library configured in `deno.importMap` for type checking, linting, and all other purposes, and a special import map for testing. 

Unfortunately, there seems to be no way to configure testing code lens to use that test-only import map, adding it as argument in `deno.codeLens.testArgs` results in an error since the import map is supplied twice.

It would be useful for cases like this to skip automatically adding `--import-map` argument if it's already present in `deno.codeLens.testArgs`, thus, allowing us to use custom import maps for testing.

zandaqo:
resolves #550

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -229,9 +229,11 @@ export function test(
     if (config.get("unstable")) {
       testArgs.push("--unstable");
     }
-    const importMap: string | undefined | null = config.get("importMap");
-    if (importMap?.trim()) {
-      testArgs.push("--import-map", importMap.trim());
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM

Really sorry for taking so long to take a look at it.

PR #571, opened 2021-11-06
fix: do not cache unsupported origins in config

kitsonk:
Ref: denoland/deno#12621

### DIFF
--- a/client/src/notification_handlers.ts
+++ b/client/src/notification_handlers.ts
@@ -8,25 +8,25 @@ export function createRegistryStateHandler(): NotificationHandler<
   RegistryStateParams
 > {
   return async function handler({ origin, suggestions }) {
-    let enable = false;
     if (suggestions) {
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
I feel like we might want to save the "No" when the user explicitly opts out by pressing "No". If they cancel though, we should do nothing. That way we won't prompt the user again about a specific registry once we eventually get around to re-fetching the manifest every 24 hours (the old extension did this).

kitsonk:
@lucacasonato you are right... good catch, I will amend.

kitsonk:
Actually that is what it does.  If you say `"No"` (specifically not `"Enable"`) it will set that origin as `false` instead of `true`, so it follows the behaviour you describe.

And denoland/deno#9931 should address the reloading of not found origins daily.

lucacasonato (approved):
Must have misread then. LGTM!

PR #593, opened 2021-12-14
docs: move docs, update/rename registry schema

kitsonk:


schemas/registry-completions.schema.json
```diff
@@ -11,9 +11,11 @@
   "properties": {
     "version": {
       "description": "The version of the schema document.",
-      "type": "number",
+      "type": "integer",
```
dsherret, line 14: I did not know there was an integer type in json schema.

schemas/registry-completions.schema.json
```diff
@@ -67,16 +69,25 @@
...
+            "https://example.com/api/packages/${package}",
+            "https://example.com/api/packages/${package}/${{version}}",
+            "/api/packages/${package}/${{version}}/${path}"
           ]
         }
       }
     }
   }
-}
+}
```
dsherret, line 93: I wonder what is removing the final newlines. Not that it really matters, but seems like we don't have code formatting on this project?

dsherret (approved):
LGTM

schemas/registry-completions.schema.json
kitsonk: I didn't either until I looked up how to constrain the value.

schemas/registry-completions.schema.json
kitsonk: Having an old formatter configured in my IDE for JSON files and not running `npm fmt` before comitting!

PR #596, opened 2021-12-17
chore: update min version

kitsonk:
Updating the minimum version of the Deno CLI.  The main reason is that Deno 1.17 ships with support for v2 of the registry suggestions, and we want to roll out a v2 version of the `deno.land/x` and `deno.land/std` registry, which would require Deno 1.17 or later to take full advantage of it.

I don't think we should publish a version of vscode_deno for a week or so, so that people organically upgrade to Deno 1.17 before we start "warning" them.


lucacasonato (approved):

lucacasonato:
Sounds like a good plan

Issue #592, opened 2021-12-13
Old references to `npm run compile` command


jespertheend:
**Describe the bug**
There's some old references to `npm run compile`, which got me confused for a bit when trying to setup my development environment.
Also, with `tsc -b` now being replaced by esbuild, does `npm run watch` still work as intended?

These are all the references to `npm run compile` I could find:

https://github.com/denoland/vscode_deno#contribute
https://github.com/denoland/vscode_deno/blob/3e5051ef6f5817461aae3c3cd62aa5e08f28e71e/package.json#L405
https://github.com/denoland/vscode_deno/blob/3e5051ef6f5817461aae3c3cd62aa5e08f28e71e/.vscode/tasks.json#L4-L15

jespertheend:
Hmm seems like `npm watch` isn't working for me.
Which makes the vscode launch task not work properly either:
https://github.com/denoland/vscode_deno/blob/3e5051ef6f5817461aae3c3cd62aa5e08f28e71e/.vscode/launch.json#L16-L19

PR #597, opened 2021-12-17
chore: use esbuild script for vscode tasks

jespertheend:
Apologies if this breaks any existing workflows.
I'm not sure what setup is currently being used to work on this extension. But `npm run watch` didn't seem to work for me.
Either way, watching seems redundant since the extension host needs to be reloaded and this already makes a build before reloading.

I've also modified the readme, a build step like `npm run compile` is no longer required as the `Launch Client` already runs the build step for you.

I've also answered a few questions I had myself in the readme.

Fixes: #592

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
       ],
       "preLaunchTask": {
         "type": "npm",
-        "script": "watch"
+        "script": "esbuild"
       }
     }
   ]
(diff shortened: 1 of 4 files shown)


jespertheend:
I don't think the workflow failed because of my changes? It runs fine in my fork: https://github.com/jespertheend/vscode_deno/runs/4642986196

bartlomieju:
Yes, GHA is having problems

PR #615, opened 2022-01-27
fix: utilize fresh arrays when muting built in ls

### LINKED ISSUE #614: `typescript-deno-plugin` re-uses instances when "muting" results
Install this extension in vscode.

https://marketplace.visualstudio.com/items?itemName=heroboy.generate-switch-case-for-enum-type

Looks like the "Generate Switch Case" menu item is never being removed. It will become more and more.

图片

kitsonk:
Fixes #614

### DIFF
--- a/typescript-deno-plugin/src/index.ts
+++ b/typescript-deno-plugin/src/index.ts
@@ -111,9 +111,7 @@ class Plugin implements ts.server.PluginModule {
         return enabled
           // we should return a new array, so if someone modifies a previous on
           // we create a new one on each call
-          ? Array.isArray(emptyReturn)
-          : []
-          ? emptyReturn
+          ? Array.isArray(emptyReturn) ? [] : emptyReturn
           : target.call(ls, ...args);
       };
     };
--- a/typescript-deno-plugin/src/index.ts
+++ b/typescript-deno-plugin/src/index.ts
@@ -108,7 +108,13 @@ class Plugin implements ts.server.PluginModule {
         const enabled = fileNameArg !== undefined
           ? this.#getSetting(args[fileNameArg] as string, "enable")
           : this.#getGlobalSettings().enable;
-        return enabled ? emptyReturn : target.call(ls, ...args);
+        return enabled
+          // we should return a new array, so if someone modifies a previous on
+          // we create a new one on each call
+          ? Array.isArray(emptyReturn)
+          : []
+          ? emptyReturn
+          : target.call(ls, ...args);
       };
     };
 
--- a/typescript-deno-plugin/src/index.ts
+++ b/typescript-deno-plugin/src/index.ts
@@ -18,7 +18,7 @@ type Parameters<T = (...args: any) => any> = T extends // deno-lint-ignore no-ex
 type CallIfDisabledFunction = <T extends ts.LanguageService, J extends keyof T>(
   fn: J,
   fileNameArg: number | undefined,
-  enabledReturn: ReturnType<T[J]>,
+  enabledReturn: (() => ReturnType<T[J]>) | ReturnType<T[J]>,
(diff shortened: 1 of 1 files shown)


heroboy:
https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin#customizing-behavior

The official wiki example modify the return object. So maybe this is not 100% safe.

kitsonk:
I'm not sure what you mean.  We aren't modifying entries.  We want "mute" the built-in language server and not call the internal method.

heroboy:
I mean `emptyReturn` can be an object, then it will be modified in the above example. 
But the current code is ok, because `emptyReturn`  can only be `undefined` or `array` now.

Another problem. But I am not 100% sure. 
```
proxy.getCompletionsAtPosition = (fileName, position, options) => {
  const prior = info.languageService.getCompletionsAtPosition(fileName, position, options);
  prior.entries = prior.entries.filter(e => e.name !== "caller");
  return prior;
};
```
The example code assume the return value of `getCompletionsAtPosition` is an object and has an array member `entries`. And your code returns `undefined`. Is it safe?

typescript-deno-plugin/src/index.ts
```diff
@@ -108,7 +108,11 @@ class Plugin implements ts.server.PluginModule {
         const enabled = fileNameArg !== undefined
           ? this.#getSetting(args[fileNameArg] as string, "enable")
           : this.#getGlobalSettings().enable;
-        return enabled ? emptyReturn : target.call(ls, ...args);
+        return enabled
+          // we should return a new array, so if someone modifies a previous on
```
dsherret, line 112:
```suggestion
          // we should return a new array, so if someone modifies a previous one
```

typescript-deno-plugin/src/index.ts
```diff
@@ -108,7 +108,11 @@ class Plugin implements ts.server.PluginModule {
...
+          // we create a new one on each call
+          ? Array.isArray(emptyReturn) ? [] : emptyReturn
```
dsherret, line 114: Not a big deal because the scope of this function is quite contained, but what might be good to do is to assert that an empty array or `undefined` is provided for this argument in order to enforce the assumptions made here.

dsherret (approved):
LGTM

kitsonk:
> The example code assume the return value of `getCompletionsAtPosition` is an object and has an array member `entries`. And your code returns `undefined`. Is it safe?

`undefined` is a valid return value for that API and `undefined` does not have instances nor is passed by reference, it is a primitive symbol. It is not passed by reference.  Therefore it is "safe".

typescript-deno-plugin/src/index.ts
kitsonk: I don't want to have a function that gets called every time, for performance reasons.  The vast majority of the values are `undefined` and invoking a function call just to get back `undefined` (or even a fresh array instance) seems like a totally unnecessary overhead to me.

As you point out the functionality is very constrained, therefore we should make the solution fit the constraints, versus making it address eventualities that will never occur.  Adding runtime type constraints doesn't make a whole lot of sense for a very specific bit of code that is very internal and doesn't change frequently.

If this were a public API to be consumed and reused in other places, maybe.

kitsonk: I looked back at the code and realised there are a few objects being passed.  So I re-worked it so that it takes either a return type value, or a function that returns a return type value and for the the objects, it is now a function that returns them.  Still very specialised, but optimised for the use case.

dsherret: I don’t think we should be concerned about the performance implications of function calls unless it can be shown that removing one has an impact on performance.

I think providing a function or undefined here would have been simpler and less error prone (ex. someone providing an array with elements in the future and not noticing what this code does), but again, not a big deal.

PR #629, opened 2022-03-01
feat: support vscode Testing API

### LINKED ISSUE #487: Support the new Testing API in VSCode 1.59+
The VSCode extension api for testing is now out of preview and part of the current shipping release (1.59, released Aug 6).

It would be great for the deno extension to support the new testing extension apis (see `createTestController` in the new `tests` namespace), it would seem to line up well with the functionality of `deno test`

https://code.visualstudio.com/updates/v1_59#_testing-apis

kitsonk:
This is a work in progress PR which complements denoland/deno#13798

It interfaces to the Deno language server to provide the Testing API to vscode, which will enabled the Testing Panel within the editor and populate it with tests that are discovered by the Deno language server.

Closes #487


client/src/testing.ts
```diff
@@ -0,0 +1,341 @@
...
+      return content.value;
+  }
+}
+
+/** Resolve the test that a test item belongs to. If the test item is a test,
+ * the result will be itself, but if the test item is a step, the result will
+ * be the owning test. */
+function getTest(testItem: vscode.TestItem): vscode.TestItem {
+  let item = testItem;
+  while (item.parent && item.parent.parent) {
```
dsherret, line 67: Is ` && item.parent.parent` there because the root is the file?

dsherret (approved):
LGTM

client/src/testing.ts
kitsonk: Correct, the test item the represents a module that contains a test will have `parent === undefined`.

PR #632, opened 2022-03-11
fix: send all workspace settings during initialisation

kitsonk:
I discovered this when working on the testing API.  Some keys were missing here, which meant that some settings weren't sent during initialisation, meaning they were stuck at their default values, even if they had been modified in the `settings.json` and wouldn't be changed unless the `settings.json` file was edited.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -24,12 +24,17 @@ const LANGUAGES = [
 /** These are keys of settings that have a scope of window or machine. */
 const workspaceSettingsKeys: Array<keyof Settings> = [
   "cache",
+  "certificateStores",
   "codeLens",
   "config",
+  "enable",
   "importMap",
   "internalDebug",
   "lint",
+  "path",
   "suggest",
+  "tlsCertificate",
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM. Good find. I wonder if there's a way we could have a single source of truth on this (maybe read from the package.json?). Otherwise maybe we could have a test that ensures the two are in sync. Maybe not worth it though.

kitsonk:
I wouldn't want to read them from the package.json at runtime, as that just seems like a runtime complication that isn't warranted.  Of course there could be a code gen solution for it, but that also seems overkill.  A test would be nice, but we currently don't have tests, which in itself is a problem, and should fix that.  All that being said, they do change infrequently, so it is sort of low risk if it is remembered.  I have added a warning to the shared type, which reflects what is in the `package.json` that we are concerned about.  Hopefully that is enough of an _aide-memoir_ when adding top level keys.

There maybe also another way to do it, just mixing in the value of `deno` config namespace into the object we send.  This was a sort of hold-over/port of the v2 stuff that really wasn't re-visited, which I guess it should.

Issue #633, opened 2022-03-11
Ability to include or exclude files to be enabled by a filter in the configuration


dsherret:
**Is your feature request related to a problem? Please describe.**

Disabling Deno for certain files currently requires setting up multiple vscode workspaces. This can be a bit extreme to setup for certain projects especially considering it takes multiple config files. It would be useful to be able to specify in the config to include or exclude certain files.

cc @kitsonk - we're going to want to get this one in or a similar solution next week (see deploy channel for more details). I believe this is just an extension side change we could make?

kitsonk:
No, it is quite a bit more complicated than that.  We should have a chat.  Because there is the enablement/disablement with the built in language server, as well as the enablement/disablement in the lsp.

This was extensively discussed in #314, which we closed with workspace folders, but then there was a lot of other discussion about it.

The protocol allows "per resource" enablement/disablement, it is just the vscode doesn't actually support per resource configuration/globbing/etc.  If we overlay something on top of it, there are challenges about how we ensure correct editor behaviour.

We should discuss further next week "face-to-face" because it isn't straightforward.

dsherret:
@kitsonk what if we had a “deno.cwd” setting so people could do `”deno.cwd”: “./backend”`? Or more complicated and if possible have the ability for the vscode plugin to understand folder scoped configuration rather than workspace scoped config?

cc @lucacasonato

kitsonk:
I would say we could experiment with `"deno.enable"` taking a `true`, `false`, or a string path, which would use the workspace root as a base. Keeping it as a simple target plus its subdirectories should help. I can work on it this week and see if I run into any issues.

dsherret:
Oh, that’s a good idea! Thanks!

kitsonk:
Hmmm... that is pretty clunky actually, because it disables the ability to edit it using the GUI.

I think `deno.enablePathOnly` would be better, where basically if it is set, the value of `deno.enable` will be ignored.  Still would use the workspace root as the base.

kitsonk:
Ok, I changed my mind again... `deno.enablePathsOnly` which takes a `string[]`. The implementation in vscode_deno is about the same, and having only one path feels like it could immediately result in needing to add the support anyways.  The UI to edit it is clean, as they provide a UI for `string[]`.  I have it working on the vscode_deno side, need to implement it in the Deno language server.

maxime1992:
Hello, does this feature supports a glob pattern?

My use case is the following:

```
| ci
[...]

PR #635, opened 2022-03-15
feat: support Deno enabling specified workspace paths

kitsonk:
Closes: #633

This PR introduces the `deno.enablePathsOnly` configuration option. When set, vscode_deno will suppress built-in TypeScript/JavaScript language service features for the enabled paths within a workspace folder.  For example, setting `"deno.enablePathsOnly": ["./worker"]` would disable diagnostics and other language service features for any files in the `worker` path of the workspace (or its sub-folders).

This will require a similar feature in the Deno language server to only provide language service features for the enabled paths.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -1,9 +1,10 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 import * as commands from "./commands";
-import { ENABLEMENT_FLAG, EXTENSION_NS } from "./constants";
+import { ENABLE_PATHS_ONLY, ENABLEMENT_FLAG, EXTENSION_NS } from "./constants";
 import { DenoTextDocumentContentProvider, SCHEME } from "./content_provider";
 import { DenoDebugConfigurationProvider } from "./debug_config_provider";
+import type { EnabledOnlyPath } from "./shared_types";
 import { DenoStatusBar } from "./status_bar";
 import { activateTaskProvider } from "./tasks";
 import { getTsApi } from "./ts_api";
@@ -28,6 +29,7 @@ const workspaceSettingsKeys: Array<keyof Settings> = [
   "codeLens",
   "config",
   "enable",
+  "enablePathsOnly",
   "importMap",
   "internalDebug",
   "lint",
@@ -41,8 +43,9 @@ const workspaceSettingsKeys: Array<keyof Settings> = [
 /** These are keys of settings that can apply to an individual resource, like
  * a file or folder. */
 const resourceSettingsKeys: Array<keyof Settings> = [
-  "enable",
   "codeLens",
+  "enable",
+  "enablePathsOnly",
 ];
 
 /** Convert a workspace configuration to `Settings` for a workspace. */
@@ -51,9 +54,6 @@ function configToWorkspaceSettings(
 ): Settings {
   const workspaceSettings = Object.create(null);
   for (const key of workspaceSettingsKeys) {
-    workspaceSettings[key] = config.get(key);
-  }
-  for (const key of resourceSettingsKeys) {
     workspaceSettings[key] = config.get(key);
   }
   return workspaceSettings;
@@ -74,6 +74,32 @@ function configToResourceSettings(
       value.defaultValue;
   }
   return resourceSettings;
+}
+
+function getEnabledOnlyPaths(): EnabledOnlyPath[] {
+  const items = [] as EnabledOnlyPath[];
+  if (!vscode.workspace.workspaceFolders) {
+    return items;
+  }
+  for (const workspaceFolder of vscode.workspace.workspaceFolders) {
+    const config = vscode.workspace.getConfiguration(
+      EXTENSION_NS,
+      workspaceFolder,
+    );
+    const enabledOnlyPaths = config.get<string[]>(ENABLE_PATHS_ONLY);
+    if (!enabledOnlyPaths || !enabledOnlyPaths.length) {
+      continue;
+    }
+    const workspace = path.normalize(workspaceFolder.uri.fsPath);
+    const paths = enabledOnlyPaths.map((folder) =>
+      path.resolve(workspace, folder)
+    );
+    items.push({
+      workspace,
+      paths,
+    });
+  }
+  return items;
 }
 
 function getWorkspaceSettings(): Settings {
@@ -103,6 +129,7 @@ function handleConfigurationChange(event: vscode.ConfigurationChangeEvent) {
         ),
       };
     }
+    extensionContext.enabledPaths = getEnabledOnlyPaths();
     extensionContext.tsApi.refresh();
     extensionContext.statusBar.refresh(extensionContext);
 
@@ -111,6 +138,11 @@ function handleConfigurationChange(event: vscode.ConfigurationChangeEvent) {
       vscode.commands.executeCommand("deno.restart");
     }
   }
+}
+
+function handleChangeWorkspaceFolders() {
+  extensionContext.enabledPaths = getEnabledOnlyPaths();
+  extensionContext.tsApi.refresh();
 }
 
 function handleDocumentOpen(...documents: vscode.TextDocument[]) {
@@ -157,6 +189,14 @@ export async function activate(
     diagnosticCollectionName: "deno",
     initializationOptions: getWorkspaceSettings(),
   };
+
+  // When a workspace folder is opened, the updates or changes to the workspace
+  // folders need to be sent to the TypeScript language service plugin
+  vscode.workspace.onDidChangeWorkspaceFolders(
+    handleChangeWorkspaceFolders,
+    extensionContext,
+    context.subscriptions,
+  );
 
   // When a document opens, the language server will query the client to
   // determine the specific configuration of a resource, we need to ensure the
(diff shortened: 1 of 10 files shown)


ry:
This option seems like something that could as well be set in the `deno.json` configuration file and be available to all editors. I'm worried that the vscode extension has as a growing number of configuration that diverge from the configuration of Deno itself.

client/src/types.d.ts
```diff
@@ -5,7 +5,11 @@ import type {
   LanguageClientOptions,
 } from "vscode-languageclient/node";
```
dsherret, line 6: > This will require a similar feature in the Deno language server to only provide language service features for the enabled paths.

So this change doesn't affect the CLI doing diagnostics for a non-enabled folder, correct? As a temporary solution, is there a way we could use the client to not send files outside that enabled folder unless they're imported?

kitsonk: > As a temporary solution, is there a way we could use the client to not send files outside that enabled folder unless they're imported? (Sorry, I could research this myself, but it's late here).

No, it gets even worse, you can't really do that.  I am working on the lsp implementation.

kitsonk:
> This option seems like something that could as well be set in the `deno.json` configuration file and be available to all editors. I'm worried that the vscode extension has as a growing number of configuration that diverge from the configuration of Deno itself.

I don't think you understand how the lsp is configured.  All these options are available in all editors.  We document them [here](https://deno.land/manual@v1.19.3/language_server/overview#settings).  In this case, and editor would like have to implement something akin to what were are doing here (suppressing built in TypeScript/JavaScript language services) to have it work cleanly.  If we limited it only to the `deno.jsonc` it would be even _more_ difficult to make it available in all editors.

Forcing people to use a `deno.jsonc` to configure their editor feels like an anti-pattern, even then, especially for a use case like this, it seems like even more problematic.  People expect to go into their editor and configure how the extensions they install behave.

There are largely two types of configuration we support for the lsp, a) expressing configuration of how Deno behaves at runtime, b) expressing configuration of how the editor behaves when evaluating Deno code.  The first category can easily be covered by a `deno.jsonc` but we currently don't allow them all to be expressed.  Even _if_ they were all expressible, it is better user experience to allow them to be expressed without requiring a `deno.jsonc` to reduce user friction.  The second category are editor specific, and in my opinion should always be configurable in the editor.

The options we support (again, available to all editors, though an client may need to be configured to allow the user to configure them):

- `deno.enable` - really an editor construct, forcing people to have a `deno.jsonc` to enable it seems silly, but auto enabling if a `deno.jsonc` is detected in the root of the workspace makes sense.  There are challenges to implement that in the editor client though, but it is possible.  It would require custom messaging from the language server to the client to support, which actually means it wouldn't be easily available in all editors.  Sort of exactly what you are arguing against.
- `deno.cache` - This is set as `DENO_DIR` in the env on CLI, not currently expressible in a `deno.jsonc`.
- `deno.certificateStores`, `deno.tlsCertificate`, `deno.unsafelyIgnoreCertificateErrors` - Mimic flags available on the command line, but not expressible in the `deno.jsonc`.  It would be good to express them in the `deno.jsonc`, and maybe remove them from the editor, but then again, they are already there, so why introduce friction in removing functionality.
- `deno.config` - Before we had auto detection, it was the only way to supply a `deno.jsonc`.  Now with auto detection, this overrides the auto detection, mimicking `--config` on the command line, which seems like something we would want to preserve.
- `deno.importMap` - This wasn't expressible in the `deno.jsonc` until very recently (will ship in Deno 1.20).  It seems illogical though to take this away from people and say "go use a `deno.jsonc`".  Especially since what you might want to use at runtime and edit time can and does very, and we yet to have an accomodation for the concept of "tasks" where you might want a different import map depending upon context.
- `deno.internalDebug` - This is a lsp only construct, logging additional information when enabled.  It has no place in a `deno.jsonc` IMO.
- `deno.codeLens.implementations`, `deno.codeLens.references`, `deno.codeLens.referencesAllFunctions`, `deno.codeLens.test`, `deno.suggest.completeFunctionCalls`, `deno.suggest.names`, `deno.suggest.paths`, `deno.suggest.autoImports`, `deno.suggest.imports.autoDiscover`, `deno.suggest.imports.hosts` - all editor specific features that should stay in the editor configruation
- `deno.lint` - a simple flag which enables or disable linting diagnostics, customisation of what diagnostics are provided is only available via the `deno.jsonc`.  This seems approrpiate.
- `deno.unstable` - Mimicks the configuration of `--unstable` on the command line and not expressible in `deno.jsonc`.  Even if it were expressible, a user might want to "override" the experience in their editor.

So back to the original purpose, `deno.enablePathsOnly` is specifically targeting what the editor does, and in many cases requires client side implementation as well, like this PR does.  It isn't a language server only thing.  Even if it was, expressing it in the `deno.jsonc` to be able to have a multi-part project seems like a lot of friction around usability of Deno.

Our current solution for vscode is that you can only use "multi-root workspaces", which are really a pain in the ass, but it meant that there was no "Deno specific" configuration, and we totally offloaded that to the editor client itself.  But that position has become untenable.  People want to have their front-end code in one subdir and their Deno code in another subdir, and not have to create a multi-root workspace, even potentially providing a template project.  This provides that in a super low friction way.

The alternative, is that when the LSP starts it scans each subdir of the workspace, and if it sees a `deno.jsonc` it "enables" it, but it comes with a few big issues:

- It is really hard to have different `deno.jsonc` configurations to apply to the same language server process.  The TypeScript language server does have a concept of "projects" and we could potentially use that, but that is a lot of integration work and change.
- It would require custom client integration and support to work, where when the language server discovers a Deno enabled subdir, it has to notify the client so that the client then suppresses any built in TypeScript/JavaScript language service features.
- How would setting the `deno.config` work then?

_Even_ if we had that auto-enablement, we would still likely want to allow users to express it in their editor config as well.

Forcing users to setup a root `deno.jsonc` and express what paths are "enabled" in their editor workspace seems non-sensical to me.

package.json
```diff
@@ -126,13 +126,27 @@
...
           "default": false,
-          "markdownDescription": "Controls if the Deno Language Server is enabled. When enabled, the extension will disable the built-in VSCode JavaScript and TypeScript language services, and will use the Deno Language Server (`deno lsp`) instead.\n\n**Not recommended to be enabled globally.**",
+          "markdownDescription": "Controls if the Deno Language Server is enabled. When enabled, the extension will disable the built-in VSCode JavaScript and TypeScript language services, and will use the Deno Language Server instead.\n\n**Not recommended to be enabled globally.**",
           "scope": "resource",
           "examples": [
             true,
             false
           ]
         },
+        "deno.enablePathsOnly": {
```
dsherret, line 136: I'm not sure about having "only" in this name as providing this setting kind of implies "only". Some other ideas in no particular order: `deno.enabledPaths` (it seems this one is used elsewhere in the code), `deno.enablePaths`, `deno.paths.enable`, `deno.include`, `deno.paths.include`?

client/src/extension.ts
```diff
@@ -76,6 +76,32 @@ function configToResourceSettings(
...
+      EXTENSION_NS,
+      workspaceFolder,
+    );
+    const enabledOnlyPaths = config.get<string[]>(ENABLE_PATHS_ONLY);
+    if (!enabledOnlyPaths || !enabledOnlyPaths.length) {
+      continue;
+    }
+    const workspace = path.normalize(workspaceFolder.uri.fsPath);
+    const paths = enabledOnlyPaths.map((folder) =>
+      path.resolve(workspace, folder)
```
dsherret, line 95: Nitpick: could we do `Uri.joinPaths(workspaceFolder.uri, folder)` here instead? I believe that handles normalization and we could avoid using the "path" module.

package.json
```diff
@@ -126,13 +126,27 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "markdownDescription": "Controls if the Deno Language Server is enabled for a specific paths only. When enabled, the extension will disable the built-in VSCode JavaScript and TypeScript language services for the specified paths, and will use the Deno Language Server instead.\n\n\n\nThe workspace folder is used as the base for the supplied paths. When a value is set, the value of `\"deno.enable\"` is ignored.\n\nIf for example you have all your Deno code in `worker` path in your workspace, you add an item with the value of `./worker`.\n\n**Not recommended to be enabled in user settings.**",
```
dsherret, line 142: "for a specific paths only" could use reworking.

dsherret, line 142: > When a value is set, the value of `\"deno.enable\"` is ignored.

It might be confusing if someone does `"deno.enable": false` and also specifies this setting.

dsherret (approved):
LGTM! Thanks for doing this.

package.json
kitsonk: I wanted to avoid confusion thinking that you needed to do `deno.enable` plus whatever paths... though I guess there is documentation to clarify that.

package.json
kitsonk: Which is why I used `only` above to make it clear.  `"deno.enable": true` can also be confusing.  The problem is doing `[ "boolean", "array" ]` as a value for `"deno.enable"` makes the UI really shitty... so breaking them out and making it clear that this overrides whatever `"deno.enable"` is set to.

client/src/extension.ts
kitsonk: Ok, I thought I had run into problems with `.fsPath` not being fully normalized, but maybe that was just me being paranoid.

ninofiliu:
Been wanting `enablePaths` for years, works like a charm, thanks guys!

PR #640, opened 2022-03-28
feat: prompt for enabling when config detected

kitsonk:
This will prompt the user to enable Deno for a workspace or workspace folder when a `deno.json` or `deno.jsonc` is configured in the root of the workspace (or workspace folder):

_Extension_Development_Host__-_deno_jsonc_—_xhr

The prompt will only occur if:

- The extension has activated.
- If a workspace is open.
- The `deno.enable` setting is not explicitly set for the workspace or workspace folder.
- A `deno.json` or `deno.jsonc` is already present in the root or is created while the workspace is open.

When a user selects either _No_ or _Enable_ the value will be set in the workspace settings (or workspace folder settings when using multi-root workspaces) and the user will no longer be prompted (unless the remove the explicit configuration setting).

### DIFF
--- a/client/src/constants.ts
+++ b/client/src/constants.ts
@@ -1,5 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
+export const ENABLE = "enable";
 export const ENABLE_PATHS = "enablePaths";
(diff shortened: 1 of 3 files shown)


dsherret (approved):
LGTM!

Issue #644, opened 2022-03-31
Deno VSCode Extension seems breaking Organize Imports


VdustR:
**To Reproduce**

1. Install / enable Deno VSCode Extension.
2. Open a node repository and open a .js or .ts file.
3. <kbd>ctrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> and select `Organize Imports`.
4. Nothing happens.

https://user-images.githubusercontent.com/29639463/160975351-f809ce6a-1613-4dd2-b651-dd9a14b0373a.mp4

**Expected behavior**

1. Uninstall / disable Deno VSCode Extension.
2. <kbd>ctrl</kbd> + <kbd>shift</kbd> + <kbd>p</kbd> and select `Organize Imports` again.
3. The imports will be organized.

https://user-images.githubusercontent.com/29639463/160975405-0e8df381-3efc-4110-bdaa-9a8dc19c698e.mp4

**Screenshots**

As described above.

**Versions**

vscode: latest

> Version: 1.66.0 (user setup)
> Commit: e18005f0f1b33c29e81d732535d8c0e47cafb0b5
> Date: 2022-03-30T05:50:14.623Z
> Electron: 17.2.0
> Chromium: 98.0.4758.109
> Node.js: 16.13.0
> V8: 9.8.177.11-electron.0
> OS: Windows_NT x64 10.0.22000

deno: n/a (node repositories)
extension: latest, v3.12.0

Maybe related: https://github.com/microsoft/vscode/issues/144203

PR #645, opened 2022-03-31
fix: do not proxy organizeImports

kitsonk:
Fixes: #644

Currently we proxy `organizeImports` for the built in tsserver.  The problem is that it doesn't really do a good job because the API doesn't take a specific file name, so we end up "muting" it, while there is no real harm in allowing this to be unmuted and we don't provide an API for it on the Deno Language Server, so we should just remove it.


lucacasonato (approved):

dsherret (approved):
LGTM

PR #666, opened 2022-05-14
Don't write configuration to settings.json if values are default

bartlomieju:


kitsonk:
I think there is a better way to accomplish this.  When you get settings in vscode, there is a way to query if there is a default value, and what you want to do is compare the set value to the default value, and only set it if it where different.

An example of determining the default value:

```ts
config.inspect("lint")?.defaultValue; // would be `false` at the moment
```

bartlomieju:
Applied the suggestion

kitsonk (approved):
LGTM

PR #683, opened 2022-06-24
fix: do not start lsp or show any messages if Deno not found and not in a Deno project

### LINKED ISSUE #678: Extension still runs even while deno is disabled
**Describe the bug**
Since I installed the Deno extension, I keep getting error notifications when I open VSCode. I had `deno` installed on another computer, and the extension got installed by settings sync. Still, I would expect that the Deno extension does not require `deno` in a workspace where `deno.enable` is `false`.

**To Reproduce**

1. Install the `denoland.vscode-deno` extension
2. Do not install `deno`
3. Set `deno.enable` to `false`
4. Open a folder with VSCode
5. See two notifications about language server and failed request

**Expected behavior**

When `deno.enable` is `false`, I expect the extension to do nothing at all, not attempt to start a language service.

**Screenshots**

image

**Versions**

vscode: 1.68.0 deno: Not installed extension: 3.12.0

dsherret:
Closes #678

### DIFF
--- a/client/src/debug_config_provider.ts
+++ b/client/src/debug_config_provider.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
 import type { Settings } from "./types";
-import { getDenoCommand } from "./util";
+import { getDenoCommandPath } from "./util";
 import * as vscode from "vscode";
 
 export class DenoDebugConfigurationProvider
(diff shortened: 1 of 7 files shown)


client/src/debug_config_provider.ts
```diff
@@ -76,7 +76,7 @@ export class DenoDebugConfigurationProvider
           type: "pwa-node",
           program: "${file}",
           env: this.#getEnv(),
-          runtimeExecutable: await getDenoCommand(),
+          runtimeExecutable: await getDenoCommandPath() ?? "deno",
```
bartlomieju, line 79: This is duplicated numerous times, maybe factor out to another helper function?

bartlomieju (approved):
LGTM, good fix

PR #711, opened 2022-08-21
Exclude demo image from extension package

kidonng:
I noticed a demo image isn't excluded from the extension package, hence this pull request. With this change, package size is down from 1.59MB to 390.76KB.

It does not break the links as VSCode rewrites assets in README to use GitHub links.


ry (approved):
Thanks!

PR #722, opened 2022-10-01
"Failed to retrieve config tasks."

lightspeedlucas:
The `deno/task` LSP request expects no parameters. Current code displays a "Failed to retrieve config tasks." error message as the request fails every time I launch if `deno.unstable` is true.


kitsonk (approved):
LGTM

PR #740, opened 2022-10-25
fix: send inlay hints config on init

kitsonk:
This fixes an issue where the inlay hint configuration is not available at startup.


bartlomieju (approved):
LGTM

PR #745, opened 2022-11-04
fix: enablePaths on Windows

### LINKED ISSUE #668: `deno.enablePaths` conflict with build in typescript language server.
In other word, `deno.enablePaths` doesn't disable default typescript language server from running specified path.

So both `deno` and `typescript` language server running on the same project...
  
 
**Versions**

```
deno 1.22.0 (release, x86_64-pc-windows-msvc)
v8 10.0.139.17
typescript 4.6.2
```

Vscode version:

```
Version: 1.67.2 (user setup)
Electron: 17.4.1
Chromium: 98.0.4758.141
Node.js: 16.13.0
V8: 9.8.177.13-electron.0
OS: Windows_NT x64 10.0.22000
```

asyncmax:
Path separators on Windows caused an incorrect behavior in `#fileNameDenoEnabled` that resulted in a non-functioning `deno.enablePaths` setting on Windows. This fixes #668.

### DIFF
--- a/typescript-deno-plugin/src/index.ts
+++ b/typescript-deno-plugin/src/index.ts
@@ -79,6 +79,9 @@ class Plugin implements ts.server.PluginModule {
 
   // determines if a specific filename is Deno enabled or not.
   #fileNameDenoEnabled(fileName: string): boolean {
+    if (process.platform === "win32") {
+      fileName = fileName.replace(/\//g, "\\");
+    }
     const settings = projectSettings.get(this.#projectName);
     if (settings?.enabledPaths) {
       const paths = settings.enabledPaths.find(({ workspace }) =>


typescript-deno-plugin/src/index.ts
```diff
@@ -79,6 +79,9 @@ class Plugin implements ts.server.PluginModule {
 
   // determines if a specific filename is Deno enabled or not.
   #fileNameDenoEnabled(fileName: string): boolean {
+    if (process.platform === "win32") {
+      fileName = fileName.replace(/\//g, "\\");
```
dsherret, line 83: I think this might break the indexing into `settings?.documents?.[fileName]` below?

dsherret (changes requested):
Good catch and thanks for the PR! This occurs because typescript normalizes all slashes as `/`.

See my comment in the PR though as I'm not sure about the fix. It might be ok.

typescript-deno-plugin/src/index.ts
dsherret: It fixes it. This is an fsPath.

webdev3301:
Hello. I have just tested v3.14.1 on Windows and the language server doesn't start.
I reversed back to v3.14.0 and everything worked fine again.

cpakken:
> Hello. I have just tested v3.14.1 on Windows and the language server doesn't start. I reversed back to v3.14.0 and everything worked fine again.

Same, enablePaths doesn't work for me either in Windows 10 (Deno v1.33.4) How do you see Deno's language server version?

cpakken:
> > Hello. I have just tested v3.14.1 on Windows and the language server doesn't start. I reversed back to v3.14.0 and everything worked fine again.
> 
> Same, enablePaths doesn't work for me either in Windows 10 (Deno v1.33.4) How do you see Deno's language server version?

deno.enablePaths seems to activate when I use the full absolute paths on windows. But the typescript server isn't disabled.
image

Here is using relative paths. Deno is not enabled
image
Works without deno.enablePaths
image

But other files in my workspace breaks

PR #758, opened 2022-12-07
Use "node" launch type instead of "pwa-node"

rotu:
Silences the warning "please use type node instead"

### DIFF
--- a/client/src/debug_config_provider.ts
+++ b/client/src/debug_config_provider.ts
@@ -39,7 +39,7 @@ export class DenoDebugConfigurationProvider
       {
         request: "launch",
         name: "Launch Program",
-        type: "pwa-node",
+        type: "node",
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks for opening the PR, I will need to test it locally before merging, but FYI I recently spoke with VSCode engineer and we'll be looking into adding dedicated "deno" config to VSCode to make debugging easier. 

Could you show where you get the error described above? I've been playing with VSCode debugger a lot recently and haven't seen this error.

rotu:
@bartlomieju screenshot attached. Looks to come from [here](https://github.com/microsoft/vscode-js-debug/blob/2627bd1b9f0f099e233d8f6ea50e186df6b01a40/src/build/generate-contributions.ts#L1068).

image

bartlomieju (approved):
LGTM, thanks @rotu

PR #762, opened 2022-12-13
feat: enable inlay hints by default (#759)

victoranderson2006:
plz let me


bartlomieju:
This has already been landed.

Issue #749, opened 2022-11-23
deno.lock should have a registered JSON Schema


kitsonk:
Currently `deno.lock` gets written out by default, but in the editor it is an "unknown" file.  It is possible that the extension can register it with a JSON schema to make it easier to view and syntax highlight the lock file.

bartlomieju:
For anyone interested in tackling this, here's a short list of what needs to be done:

1. `package.json` should be updated in the `jsonValidation` field, to add:
```json
    {
        "fileMatch": [
          "deno.lock",
        ],
        "url": "./schemas/lockfile.schema.json"
    },
```
2. A new file "./schemas/lockfile.schema.json" should be added that describes the structure of the lock file.
3. That would probably be enough to get syntax highlighting working.

A nice addition would be to make `deno.lock` have a Deno icon, similar to `deno.json`

bartlomieju:
I think I might be have been wrong and it should actually use `"languages"` mapping:
```json
    "languages": [{
      "id": "json",
      "aliases": ["json"],
      "extensions": [".lock"]
    }],
```

LeoDog896:
All good! Attempted to implement this with #766

PR #766, opened 2022-12-17
Add lockfile schema

LeoDog896:
This adds support for `deno.lock` with autocomplete and JSON recognition

This also solves #749

### DIFF
--- a/schemas/lockfile.schema.json
+++ b/schemas/lockfile.schema.json
@@ -55,7 +55,7 @@
             "properties": {
               "integrity": {
                 "type": "string",
-                "description": "The hash of the npm package itself as well as the name of the hash.",
+                "description": "The hash of the npm package itself as well as the name of the hash."
               },
               "dependencies": {
                 "type": "object",
--- a/schemas/lockfile.schema.json
+++ b/schemas/lockfile.schema.json
@@ -16,6 +16,7 @@
     "version": {
       "description": "The version of the lockfile.",
       "type": "string",
+      "default": "2",
       "enum": [
         "2"
       ]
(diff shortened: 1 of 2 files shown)


schemas/lockfile.schema.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/denoland/vscode_deno/blob/main/schemas/lockfile.schema.json",
+  "title": "Deno Lock file",
+  "description": "An import map which is used to remap imports when modules are loaded.\n\nSee https://deno.land/manual@v1.6.0/linking_to_external_code/import_maps in the Deno documentation for more information.",
+  "type": "object",
+  "required": [
+    "version",
+    "remote"
+  ],
```
bartlomieju, line 10: This file needs to be updated

LeoDog896: Fixed description of both the root and the sub-property, version

schemas/lockfile.schema.json
```diff
@@ -0,0 +1,43 @@
...
+      ]
+    },
+    "remote": {
+      "description": "A map of remote URLs to their respective hashes.",
+      "type": "object",
+      "additionalProperties": {
+        "type": "string",
+        "description": "The hash as the value for the remote URL as the key."
+      }
+    }
```
bartlomieju, line 30: It's missing `npm` property that is used to store integrities of `npm` packages. You can see example lock file here: https://github.com/denoland/deno/blob/7b212bc5741ea0cfb6bd87c280be0c864cb32813/cli/tests/testdata/npm/lock_file/lock.json

bartlomieju (approved):
LGTM, thanks for looking into this @LeoDog896!

LeoDog896:
By the way, this does close #749

PR #608, opened 2022-01-13
Update default debug configuration

RomanKornev:
I've updated the debug config generation according to the manual at https://deno.land/manual@v1.12.2/getting_started/debugging_your_code since the autogenerated config was not working.
Fixes #476.

PR #775, opened 2022-12-26
feat: make `--inspect-wait` default when debugging

### LINKED ISSUE #774: Debugger: should `--inspect-wait` be default?
### Discussed in https://github.com/denoland/vscode_deno/discussions/772

<div type='discussions-op-text'>

<sup>Originally posted by **Leokuma** December 25, 2022</sup>
With [this addition](https://github.com/denoland/deno/pull/17001), I wonder if it would make sense to make `--inspect-wait` default instead of `--inspect` when debugging so that no breakpoints are ever missed.</div>

Leokuma:
Fixes #774 and possibly makes #608 obsolete.

Please let me know if I should change or test something specific. I've tested with current and old Deno versions.

I had to change the argument received in the constructor because I needed to check the version of the Deno Server.

From what I observed (code below), `getSettings === extensionContext.clientOptions.initializationOptions`, so I adjusted the calls.

https://github.com/denoland/vscode_deno/blob/5ab30d8c7e966add3db6a4ecef1cfccd49cf0d3b/client/src/extension.ts#L198

https://github.com/denoland/vscode_deno/blob/5ab30d8c7e966add3db6a4ecef1cfccd49cf0d3b/client/src/extension.ts#L245

### DIFF
--- a/client/src/debug_config_provider.ts
+++ b/client/src/debug_config_provider.ts
@@ -1,21 +1,22 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-import type { Settings } from "./types";
+import type { DenoExtensionContext } from "./types";
 import { getDenoCommandName } from "./util";
+import * as semver from "semver";
 import * as vscode from "vscode";
 
 export class DenoDebugConfigurationProvider
   implements vscode.DebugConfigurationProvider {
-  #getSettings: () => Settings;
+  #extensionContext: DenoExtensionContext;
 
   #getEnv() {
-    const cache = this.#getSettings().cache;
+    const cache = this.#extensionContext.clientOptions.initializationOptions().cache;
     return cache ? { "DENO_DIR": cache } : undefined;
   }
 
   #getAdditionalRuntimeArgs() {
     const args: string[] = [];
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks

client/src/debug_config_provider.ts
```diff
@@ -30,8 +31,18 @@ export class DenoDebugConfigurationProvider
...
 
-  constructor(getSettings: () => Settings) {
-    this.#getSettings = getSettings;
+  #getInspectArg() {
+    const version = this.#extensionContext.serverInfo?.version;
+
+    if (version && semver.valid(version) && semver.satisfies(version, ">=1.29.0")) {
+      return "--inspect-wait"
+    } else {
+      return "--inspect";
```
bartlomieju, line 40: Now that I'm thinking about it... we should probably default to `--inspect-brk` in older versions of Deno, otherwise some breakpoints might not get registered and hit. @Leokuma could you change that?

Leokuma: Sure! New PR incoming.

Leokuma: #779

PR #779, opened 2022-12-28
feat: make `--inspect-brk` default in older versions of Deno

Leokuma:
As requested https://github.com/denoland/vscode_deno/pull/775#discussion_r1057971783


bartlomieju (approved):
LGTM, many thanks @Leokuma

PR #790, opened 2023-01-15
feat: add default formatter question on initialize

### LINKED ISSUE #789: Deno.fmt seems not to be executing by default
**Describe the bug**

This is both a bug and a feature request (which I'm willing to implement if needed), but when setting up the workspace, `deno.fmt` is enabled, but even with all the extensions disabled and just deno enabled, it doesn't seem to correctly set the default formatter to `denoland.vscode_deno` by default.

**To Reproduce**

To get a better reproducibility disable all your extensions by `cmd+shif+p` -> `Disable all installed extensions`:

image

Then enable only the `deno` extension on the extension manager.

1. Set up a new directory anyywhere
2. Initialize deno workspace, answer yes to all questions
3. There should be a `.vscode/settings.json` file with these contents:

```json
{
  "deno.enable": true,
  "deno.unstable": true,
}
```

4. Create a `*.ts` file anywhere, type any full TS command with a different formatting then Deno expects, like:

```ts
import { resolve } from 'https://deno.land/std/path/mod.ts'
// 									 /\ single quotes								  /\ missing ;
```

5. Save the file, if you haven't configured `editor.formatOnSave` option, this shouldn't do anything by default, however if you do `cmd+shift+p` -> `format document`, is also doesn't do anything.

**Expected behavior**

It should apply `deno.fmt` on the file and the line should look like:

```ts
import { resolve } from "https://deno.land/std/path/mod.ts";
//									  /\ double										   /\ correct ;
```

**Screenshots**

This is a gif of the current behaviour:

Kap Recording - 2023-01-15 12 12 14

I figured out that, if you set `"[typescript]": { "editor.defaultFormatter": "denoland.vscode_deno" }` in the settings, this would work, like so:

deno-true

**Versions**

vscode: 1.74.3 deno: 1.29.2 extension: v3.16.0

## Proposed solution

Include, by default, these lines in the settings when setting up the workspace:

```json
{
  "[typescript]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "[typescriptreact]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  }
}
```

Also would be a nice addition to ask if the user wants to format on save and then include `editor.formatOnSave` as `true`.

khaosdoctor:
This is a proposed solution to #789.

fixes #789


client/src/commands.ts
```diff
@@ -78,6 +78,7 @@ export function initializeWorkspace(
     try {
       const settings = await pickInitWorkspace();
       const config = vscode.workspace.getConfiguration(EXTENSION_NS);
+      const languageSpecificConfigs = await Promise.all(["typescript", "typescriptreact"].map(async (languageId) => vscode.workspace.getConfiguration("", { languageId })))
```
khaosdoctor, line 81: **Note**: Did this because, since language specifics are not namespaced by the extension settings, we couldn't really rely on the other variable since it would return nothing. So I added it as an array in case we need to add more support (JS, JSX) in the future, so this would be very easy to do.

Also, in case we need to change any other language-specific configs in the future we just need to iterate through it again.

bartlomieju:
Thanks for the PR @khaosdoctor, I modified it a bit. Fresh started adding these settings by default in https://github.com/denoland/fresh/pull/1431 so I think it totally makes sense to add it.

bartlomieju (commented):
Thanks for the PR @khaosdoctor, I modified it a bit. Fresh started adding these settings by default in https://github.com/denoland/fresh/pull/1431 so I think it totally makes sense to add it.

@dsherret are you fine with this PR?

khaosdoctor:
Yay! Thanks for merging!

devlopersabbir (approved):

PR #798, opened 2023-01-28
fix: run only the clicked test via code lens

### LINKED ISSUE #554: Running Deno tests in the editor should not use fuzzy test name matching
**Describe the bug**

Testing a `Deno.test` runs multiple tests with the same prefix

**To Reproduce**

```ts
import { assertEquals } from "https://deno.land/std@0.109.0/testing/asserts.ts";

Deno.test({
  name: "foo",
  fn: () => {
    assertEquals(true, true);
  },
});

Deno.test({
  name: "foo bar",
  fn: () => {
    assertEquals(true, true);
  },
});

Deno.test({
  name: "foo bar baz",
  fn: () => {
    assertEquals(true, true);
  },
});
```

**Expected behavior**

When the "Run Test" button is clicked, I expect only that test to run and no other.

**Screenshots**

**Versions**

deno: 1.14.2 extension: 3.9.1

Leokuma:
Fixes #554 by applying [this suggestion](https://github.com/denoland/vscode_deno/issues/554#issuecomment-932867639).

The regex escaping function is [taken from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping).


bartlomieju (commented):
Thanks for the PR @Leokuma, this is a bad papercut so I'm happy to land it.

Could you add some unit tests that asserted escaped argument value? I assume we will be passing something like `deno test -A --filter /exact_test_case_name/ ./my_test.ts`. Is that correct?

Leokuma:
@bartlomieju That's right. More specifically, the transformation will be like this:

`Test Foo` > `/^Test Foo$/`

`^` and `$` are used to prevent partial matches.

Let me know if you would like me to test specific cases.

I would be glad to add unit tests, but how should we do it? Should we choose a Node testing lib?

bartlomieju:
@Leokuma sorry for a slow response.

> ^ and $ are used to prevent partial matches.

Sounds good, it was not obvious for me at first look.

> I would be glad to add unit tests, but how should we do it? Should we choose a Node testing lib?

Ideally we wouldn't take any additional dependencies. We probably would need to update to latest Node version on CI, so we could use its built-in testing functionality (https://nodejs.org/api/test.html).

Is it something you could try to tackle (preferably in a separate PR) by updating the `.github/workflow/ci.yml` file to use Node v20, or should I look into it?

nayeemrmn (approved):
LGTM!

Cargo has `cargo test some_test -- --exact` to toggle exact matching. I wonder if we should add `deno test --filter-exact some_test` and use it here?

Leokuma:
OMG I forgot about this. Yes I can try to create unit tests, just keep in mind that I've never done that in Node before 😅. I've been very busy lately but hopefully I'll have some free time next week.

@nayeemrmn I'm in favor of your suggestion because maybe "exact filtering" is a too trivial use case to be possible only through regex "trickery".

PR #799, opened 2023-01-28
fix: respect user's answers for enabling or disabling lint and the unstable flag during workspace initialization

### LINKED ISSUE #786: Even if I select "lint:Yes" in "Deno: initializeWorkspace", lint is not enabled.
After #665, the "deno.lint" setting should be true by default, but it doesn't seem to be the case.

Leokuma:
Fixes #786.


bartlomieju (approved):
Thanks

client/src/commands.ts
```diff
@@ -84,13 +84,9 @@ export function initializeWorkspace(
       assert(lintInspect);
       const unstableInspect = config.inspect("unstable");
       assert(unstableInspect);
-      // Don't write settings if they have default values.
```
dsherret, line 87: This was initially specifically done to prevent the .vscode/settings.json file from having unnecessary information. Probably we should revert this or change the question or remove the question instead.

Leokuma: > Bartek brought up the point that someone may have a global setting that differs from the default.

As far as I remember, that was actually the main point.

Also, the user may change the global setting at some point after initializing the Deno workspace. Then in that case the workspace config would be "overwritten" if it's not persisted.

But if you guys decide otherwise, just let know and I can change it.

dsherret: This is good. No need to change anything. Thanks!

Issue #826, opened 2023-03-18
Feature Request: Cache On Save


ayame113:
I wish the `deno cache` command was executed when I hit `ctrl+s`.

This was previously implemented in the Deno plugin for the Atom editor and was very useful.

PR #831, opened 2023-03-31
feat: cache on save

CGQAQ:
This PR added a feature called cache on save that includes:

- A setting entry `deno.cacheOnSave` of type of bool, defaults to `true`
- A listener of `onDidSaveTextDocument` that will call `deno.cache` command if `deno.cacheOnSave` is set to `true`

closes: #826

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -182,6 +182,17 @@ function handleTextDocumentSave(doc: vscode.TextDocument) {
     return;
   }
   if (extensionContext.workspaceSettings.cacheOnSave) {
+    const diagnostics = vscode.languages.getDiagnostics(doc.uri);
+    if (
+      !diagnostics.some((it) =>
+        it.code === "no-cache" || it.code === "no-cache-npm"
+      )
+    ) {
+      console.log("no-cache diagnostic not found, skipping cache...");
+      return;
+    }
+
+    console.log("no-cache diagnostic found, caching...");
     vscode.commands.executeCommand("deno.cache");
   }
 }
@@ -247,7 +258,7 @@ export async function activate(
     handleTextDocumentSave,
     extensionContext,
     context.subscriptions,
-  )
+  );
 
   extensionContext.statusBar = new DenoStatusBar();
   context.subscriptions.push(extensionContext.statusBar);
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -188,7 +188,6 @@ function handleTextDocumentSave(doc: vscode.TextDocument) {
         it.code === "no-cache" || it.code === "no-cache-npm"
       )
     ) {
-      console.log("no-cache diagnostic not found, skipping cache...");
       return;
     }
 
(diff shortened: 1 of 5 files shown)


CGQAQ:
ping @lucacasonato @crowlKats

CGQAQ:
ping @bartlomieju

client/src/extension.ts
```diff
@@ -171,6 +172,15 @@ function handleDocumentOpen(...documents: vscode.TextDocument[]) {
   }
 }
 
+function handleTextDocumentSave(doc: vscode.TextDocument) {
+  if (!LANGUAGES.includes(doc.languageId)) {
+    return;
+  }
+  if (extensionContext.workspaceSettings.cacheOnSave) {
+    vscode.commands.executeCommand("deno.cache");
```
dsherret, line 180: Is there a way to do this only if deno says there's stuff to cache in the current file? (ex. if there are diagnostics or cache quick fixes). This can be quite expensive.

CGQAQ: > Is there a way to do this only if deno says there's stuff to cache in the current file

I think so, will take a look soon...

CGQAQ:
@dsherret pls take a look again

client/src/extension.ts
```diff
@@ -176,6 +177,26 @@ function handleDocumentOpen(...documents: vscode.TextDocument[]) {
...
+    return;
+  }
+  if (extensionContext.workspaceSettings.cacheOnSave) {
+    const diagnostics = vscode.languages.getDiagnostics(doc.uri);
+    if (
+      !diagnostics.some((it) =>
+        it.code === "no-cache" || it.code === "no-cache-npm"
+      )
+    ) {
+      console.log("no-cache diagnostic not found, skipping cache...");
```
dsherret, line 191:
```suggestion
```

This might be a little noisy. Let's remove it.

package.json
```diff
@@ -133,6 +133,16 @@
             false
           ]
         },
+        "deno.cacheOnSave": {
+          "type": "boolean",
+          "default": true,
```
dsherret, line 138:
```suggestion
          "default": false,
```

Let's make this false to start and get feedback on.

typescript-deno-plugin/src/index.ts
```diff
@@ -29,6 +29,7 @@ const projectSettings = new Map<string, PluginSettings>();
  * received from the extension. */
 const defaultSettings: Settings = {
   cache: null,
+  cacheOnSave: true,
```
dsherret, line 32:
```suggestion
  cacheOnSave: false,
```

package.json
```diff
@@ -133,6 +133,16 @@
             false
           ]
         },
+        "deno.cacheOnSave": {
```
dsherret, line 136: Would you be able to update README.md as well? There is an outline of all the config there.

dsherret (changes requested):
Let's make this `false` by default so we can try it out a bit more in the wild and get feedback on it.

dsherret (approved):
LGTM. Thanks!

Issue #837, opened 2023-04-13
Cannot import npm specifier modules due to firewall


justinmchase:
**Describe the bug**
I cannot configure `NPM_CONFIG_REGISTRY` in the vscode extension.

https://deno.com/blog/v1.29#custom-registry-support-via-environment-variable

**To Reproduce**

1. Connect to a corporate VPN which blocks npm
2. Setup an Artifactory proxy
3. Attempt to import something with an npm specifier
```ts
import * as express from "npm:express";
```

**Expected behavior**

I expect to be able to have this succeed by importing through the proxy.

I would imagine it could be as simple as adding an option `npmRegistry` to the settings which would set that env variable before invoking deno.

```json
{
  "deno.npmRegistry": "<url>"
}
```

**Screenshots**

![Screen Shot 2023-04-13 at 9 38 02 AM](https://user-images.githubusercontent.com/10974/231794588-68583ee5-4e31-4650-89b3-314f0e7d6ea2.png)

![Screen Shot 2023-04-13 at 9 37 55 AM](https://user-images.githubusercontent.com/10974/231794620-c46ac0df-eca3-4751-a696-ae2e3ccbde77.png)

**Versions**

vscode: ` 1.67.2` deno: `1.32.4` extension: `v3.17.0`

oscar6echo:
I reach a similar wall.

Config: vscode: `1.89.1`, deno: `1.43.3`, extension: `3.37.1`

Behind a corp proxy I can use deno CLI with [specific env variables](https://docs.deno.com/runtime/manual/basics/env_variables#special-environment-variables):
+  `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` 
+ `DENO_CERT` to use certificates including corp ones

But vscode deno extension does not seem aware of proxy env variables. As a result remote links appear underlined in red wiggle and "Uncached or missing remote URL" on hover e.g.:

![image](https://github.com/denoland/vscode_deno/assets/2219383/3614433a-f230-42bc-810e-5785eb6bbaeb)

![image](https://github.com/denoland/vscode_deno/assets/2219383/8ac8a6b8-9c41-4d24-aff0-e3fc870fa46a)

_NOTE_: Here they look ok as I write from **outside corp env** 

An unfortunate consequence for dev exp is that all objects from these outside links are considered `any` which breaks type checks and forces to use `// deno-lint-ignore no-explicit-any` in the rest of the code.

[...]

PR #858, opened 2023-05-30
add "registry" Settings option

marcushultman:
Makes the LSP setting added in https://github.com/denoland/deno/pull/19317 configurable in the VS Code extension

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -41,6 +41,7 @@ const workspaceSettingsKeys: Array<keyof Settings> = [
   "inlayHints",
   "internalDebug",
   "lint",
+  "npmRegistry",
   "path",
   "suggest",
   "testing",
--- a/client/src/shared_types.d.ts
+++ b/client/src/shared_types.d.ts
@@ -65,6 +65,8 @@ export interface Settings {
   internalDebug: boolean;
   /** Determine if the extension should be providing linting diagnostics. */
   lint: boolean;
+  /** A custom npm registry used for npm specifiers. */
+  npmRegistry: string | null;
   /** Specify an explicit path to the `deno` binary. */
   path: string | null;
   suggest: {
--- a/package.json
+++ b/package.json
@@ -157,6 +157,15 @@
             "C:\\Program Files\\deno\\deno.exe"
           ]
         },
+        "deno.npmRegistry": {
+          "type": "string",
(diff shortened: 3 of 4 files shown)


bartlomieju (commented):
Let's use `npmRegistry` as suggested in https://github.com/denoland/vscode_deno/issues/837.

While this set ups settings nicely I think we need to still figure out integration with `deno` CLI itself.

marcushultman:
@bartlomieju I pivoted in https://github.com/denoland/deno/pull/19317 to read `npmRegistry` from the configuration file instead of having the option only in the LSP (to be used in the LSP as well as deno processes spawned inside and outside of VS Code). There were some opinions that it should remain an LSP-only option, which I think might confuse things where the editor is happy but CLI invocations complain. I ended up changing my workflow by launching VS Code with the `NPM_CONFIG_REGISTRY` env var set, so it would apply to both the LSP and processes spawned from the terminal in the editor. Right now both of these PRs are in stand-still.

bartlomieju:
> @bartlomieju I pivoted in [denoland/deno#19317](https://github.com/denoland/deno/pull/19317) to read `npmRegistry` from the configuration file instead of having the option only in the LSP (to be used in the LSP as well as deno processes spawned inside and outside of VS Code). There were some opinions that it should remain an LSP-only option, which I think might confuse things where the editor is happy but CLI invocations complain. I ended up changing my workflow by launching VS Code with the `NPM_CONFIG_REGISTRY` env var set, so it would apply to both the LSP and processes spawned from the terminal in the editor. Right now both of these PRs are in stand-still.

Hey @marcushultman, sorry. for a slow response. I discussed this PR with @dsherret and we're not sure if that's the right approach to use `npmRegistry` from the config file. I don't remember the specifics but I believe David preferred to go with the env var approach. Let me get back to you on this one.

dsherret:
I think a `"deno.npmRegistry"` configuration would be nice that sets the NPM_REGISTRY_CONFIG environment variable (https://github.com/denoland/vscode_deno/blob/bce5713084eebb789e8257fa23e102cc6cc3c2e8/client/src/commands.ts#L146). It would require no CLI changes.

nayeemrmn:
This can be done with `deno.env` and `NPM_REGISTRY_CONFIG` now.

PR #861, opened 2023-06-06
feat: show output panel on status bar click

marvinhagemeister:
This PR adds a click handler to the status bar item which opens the output panel. Noticed that the prettier extension has this too and it's kinda neat from a UX perspective for debugging. Otherwise you'd need to open the output panel manually and pick the appropriate extension from the dropdown menu.

Screenshot 2023-06-06 at 11 05 32


bartlomieju (approved):
LGTM, great idea

PR #886, opened 2023-08-01
Enable by default for workspaces with deno.json{c}

### LINKED ISSUE #880: Auto-enable extension when deno.json/deno.jsonc is present
It'd be awesome if the extension could by default look for a deno.json or deno.jsonc file and if one of the two is available, set `deno.enable` automatically to `true`.

**Why?**

A lot of Deno projects (e.g. the std library) don't configure `deno.enable` for just the workspace, making it annoying for people who don't have  `deno.enable` set to `true` globally.

Also, this would allow us to have one file less in our codebases. :)

felipecrs:
WIP, please don't review yet.

Fixes #880

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -32,7 +32,7 @@ import type {
   Location,
   Position,
 } from "vscode-languageclient/node";
-import { getWorkspacesEnabledInfo } from "./enable";
+import { isEnabled } from "./enable";
 
 // deno-lint-ignore no-explicit-any
 export type Callback = (...args: any[]) => unknown;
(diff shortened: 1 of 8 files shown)


bartlomieju:
Hey @felipecrs, how's the PR going? I wanted to give you a heads up that we're working on support "workspaces" that will require some changes to logic of `vscode_deno` that might overlap with your PR. If you don't feel like finishing it, that's okay, we should get back to it soon enough.

felipecrs:
Oh, thanks for the heads up. I gave it a try but I got stuck at the initialization part. I suspect I'm not making correct use of async/await during the extension initialization.

Then, I just forgot. lol

Honestly I don't feel like I'm getting back to it. But feel free to reuse the code from the PR, if it's of any help.

Good luck!

PR #893, opened 2023-08-19
feat: allow --inspect-wait when debugging test

jeiea:
### Motivation

vscode_deno offers a code lens action "Run Test | Debug". However when clicking the "Debug", it triggers a breakpoint which an user did not set.

### Changes made

- Unified the logic for selecting the appropriate debugging flag (`--inspect-brk` or `--inspect-wait`) based on the behavior of pull request #775.

### Testing

Tested on windows, not sure about other OS but I think there should be no problem.

### Notes for Reviewers

The breakpoint continues to be triggered only on the first time and not from the second. I'm not sure about underlying cause.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -19,7 +19,12 @@ import * as tasks from "./tasks";
 import { DenoTestController, TestingFeature } from "./testing";
 import type { DenoExtensionContext, TestCommandOptions } from "./types";
 import { WelcomePanel } from "./welcome";
-import { assert, getDenoCommandName, getDenoCommandPath } from "./util";
+import {
+  assert,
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
@nayeemrmn please review

nayeemrmn (approved):
LGTM

> The breakpoint continues to be triggered only on the first time and not from the second. I'm not sure about underlying cause.

That's not happening for me, but I am seeing after the first run that VSCode defaults to having the debug console open but after future runs it defaults to the terminal where you can see the test results. Maybe that's what you're seeing? Either way tests are being run from the first time, so looks good.

PR #902, opened 2023-08-31
feat: enable via config file detection

### LINKED ISSUE #880: Auto-enable extension when deno.json/deno.jsonc is present

nayeemrmn:
Closes #880.
Needs https://github.com/denoland/deno/pull/20334.

### DIFF
--- a/client/src/enable.ts
+++ b/client/src/enable.ts
@@ -3,6 +3,7 @@
 import { ENABLE, ENABLE_PATHS, EXTENSION_NS } from "./constants";
 
 import * as vscode from "vscode";
+import { DenoExtensionContext } from "./types";
 
 export interface WorkspaceEnabledInfo {
   folder: vscode.WorkspaceFolder;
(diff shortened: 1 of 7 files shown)


nayeemrmn:
https://github.com/denoland/vscode_deno/assets/29990554/36a1d263-2487-4316-87a7-3de417635d42

bartlomieju (commented):
Is this change backwards compatible? Ie. after landing and releasing extension will older versions of Deno (<= 1.36.x) still work?

nayeemrmn:
> Is this change backwards compatible? Ie. after landing and releasing extension will older versions of Deno (<= 1.36.x) still work?

The protocol is. However, with a new extension version; an old Deno version; `deno.enable` unspecified and `deno.json` in the directory, both the Deno and Node/TS language servers will be inactive. We expect such users to have had `deno.enable: true` set though. So rare edgecase with not so bad behaviour.

nayeemrmn:
Actually wait, I checked and there is a problem with null not being accepted for `enable` for an old lsp version. I thought the `#[serde(default)]` above it handled that...

nayeemrmn:
Okay it's because https://github.com/denoland/deno/blob/9d58c896dc11be39727729b74ebe5d5d7bb58767/cli/lsp/config.rs#L239 doesn't have a `#[serde(default)]` above it like https://github.com/denoland/deno/blob/9d58c896dc11be39727729b74ebe5d5d7bb58767/cli/lsp/config.rs#L292 does.

bartlomieju:
Alright, so a new extension version will cause problems? Maybe we can make this feature conditional depending on the Deno binary version and only use it for >= v1.37?

bartlomieju (approved):
LGTM 🚀

nayeemrmn:
> Alright, so a new extension version will cause problems? Maybe we can make this feature conditional depending on the Deno binary version and only use it for >= v1.37?

Done. For anyone that wants to play around with it, remove the check from https://github.com/denoland/vscode_deno/pull/902/commits/b6a0d19ca12f0439e12d3387de61d7d693431697.

PR #905, opened 2023-09-05
fix: check for existing commands before registering them

nayeemrmn:
For https://github.com/denoland/deno/pull/20111.

Also includes `"enableBuiltinCommands": true` in the server init options. As a temporary restraint to avoid breakage, the LSP will not register built-in commands unless it sees that.


bartlomieju (approved):
LGTM, please adjust the description before landing; "enableBuiltinModules" threw me off 😅

PR #930, opened 2023-09-21
fix: regression when caching via the command palette

nayeemrmn:
When moving the `deno.cache` command server-side (https://github.com/denoland/deno/pull/20111), it stopped working when you run it from the command palette since no args are passed in that case. I was testing it with the code action only. This separates the one intended to be run that way into a separate command: `deno.client.cacheActiveDocument`. There's no sense in implementing these under the same command.

For clarity and future-compat, this renames all of the commands that are supposed to be implemented by the client/editor from `deno.<name>` to `deno.client.<name>`.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -184,7 +184,7 @@ function handleConfigurationChange(event: vscode.ConfigurationChangeEvent) {
       event.affectsConfiguration("deno.path") ||
       event.affectsConfiguration("deno.maxTsServerMemory")
     ) {
-      vscode.commands.executeCommand("deno.restart");
+      vscode.commands.executeCommand("deno.client.restart");
     }
   }
 }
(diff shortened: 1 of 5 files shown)


client/src/extension.ts
```diff
@@ -184,7 +184,7 @@ function handleConfigurationChange(event: vscode.ConfigurationChangeEvent) {
       event.affectsConfiguration("deno.path") ||
       event.affectsConfiguration("deno.maxTsServerMemory")
     ) {
-      vscode.commands.executeCommand("deno.restart");
+      vscode.commands.executeCommand("deno.client.restart");
```
bartlomieju, line 187: This is not gonna be backwards compatible; can we make it backwards compatible for 1-3 months?

bartlomieju: Err, this is only `vscode_deno`. Doesn't have to do anything with the LSP.

bartlomieju (approved):

PR #937, opened 2023-09-25
chore: version check for client-side "deno.cacheOnSave"

nayeemrmn:
https://github.com/denoland/deno/pull/20632 moves `deno.cacheOnSave` handling to the server-side, available as of `1.37.0+33f8432`. Disables the implementation here for LSP versions > 1.37.0 (this leaves a dead-zone where you won't have either for some stale canary versions, but that's okay and we want to ship this to proper nightly users).

### DIFF
--- a/typescript-deno-plugin/src/index.ts
+++ b/typescript-deno-plugin/src/index.ts
@@ -143,7 +143,6 @@ class Plugin implements ts.server.PluginModule {
       // deno-lint-ignore no-explicit-any
       const target = (ls as any)[fn];
       return (...args) => {
-        this.#log(fn, args);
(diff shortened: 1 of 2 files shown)


typescript-deno-plugin/src/index.ts
```diff
@@ -143,6 +143,7 @@ class Plugin implements ts.server.PluginModule {
       // deno-lint-ignore no-explicit-any
       const target = (ls as any)[fn];
       return (...args) => {
+        this.#log(fn, args);
```
bartlomieju, line 146: Stray log?

bartlomieju (approved):
LGTM

PR #949, opened 2023-09-28
feat: register notebook cells with the language client

nayeemrmn:
Start sending notebook cells to the LSP.

VSCode stores these cells as documents with URLs like:
```
vscode-notebook-cell:/home/nayeem/projects/vscode_deno/notebook.ipynb#W0aIKpasdpH53D%3D
```
When sending them to the LSP, I've mapped them to:
```
deno-file-fragment:/home/nayeem/projects/vscode_deno/notebook.ipynb#W0aIKpasdpH53D%3D
```
because it's neutral to other editors and because the handling in the LSP is pretty generic wrt 'file fragments'. Neither `file:///` (actual FS files with no fragment) nor `deno:` (static virtual documents with paths not related to the actual FS) are appropriate for this.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -256,6 +256,7 @@ export async function activate(
 ): Promise<void> {
   extensionContext.outputChannel = extensionContext.outputChannel ??
     vscode.window.createOutputChannel(LANGUAGE_CLIENT_NAME);
+  const p2cMap = new Map<string, string>();
   extensionContext.clientOptions = {
     documentSelector: [
       { scheme: "file", language: "javascript" },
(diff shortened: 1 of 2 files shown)


client/src/extension.ts
```diff
@@ -269,7 +270,30 @@ export async function activate(
...
+      { notebook: "*", language: "javascript" },
+      { notebook: "*", language: "javascriptreact" },
+      { notebook: "*", language: "typescript" },
+      { notebook: "*", language: "typescriptreact" },
     ],
+    uriConverters: {
+      code2Protocol: (uri) => {
+        if (uri.scheme == "vscode-notebook-cell") {
+          const string = uri.with({
+            scheme: "deno-file-fragment",
```
dsherret, line 282: 👍

dsherret (approved):
Sweet! Are we going to be able to tell the order server side?

nayeemrmn:
> Sweet! Are we going to be able to tell the order server side?

No they're stored as unrelated documents in my current mvp. (But now that I think about it, if we support that I don't think a generic `deno-file-fragment` scheme makes sense... might have to make it `deno-notebook-cell`.)

The fragment of the URI seems to contain an index to the cell. Other than that I'm not sure how to decode it properly.

PR #952, opened 2023-10-02
feat(fresh): add fresh route view, snippets

hashrock:
This PR introduces Fresh features:

# Route view

image

The Route view is designed as a live documentation, especially for beginners of Fresh.

- Special files like `_app.tsx` are displayed with distinct icons and are accompanied by links to the documentation. This prevents beginners from being confused when they encounter these files right after initializing a project.
- URL Tester for testing dynamic routes.
- Parameters for dynamic routes can be entered directly into a text input, and can be previewed in a browser with the icon on the right.

**Constraints:**

Users need to start the Fresh server before the browser preview. Also, the URL is fixed at http://localhost:8000/.

**Note:**

The Fresh routes view should not appear in non-Fresh projects as it checks `fresh.gen.ts`. I believe I haven't changed the behavior of the existing Deno extension, but please let me know if there are any oversights.

# Code template

image

image

# Code snippets

image

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -19,6 +19,7 @@ import { getTsApi } from "./ts_api";
 import type { DenoExtensionContext, Settings } from "./types";
 import { assert } from "./util";
 import * as util from "util";
+import * as freshExtension from "./fresh/extension";
 
 import * as vscode from "vscode";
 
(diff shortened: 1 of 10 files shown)


nayeemrmn (approved):
LGTM!

PR #954, opened 2023-10-03
fix: increase client shutdown timeout from 2s to 10s

nayeemrmn:
There's a big chunk of blocking sync IO we do on the main LSP thread at https://github.com/denoland/deno/blob/v1.37.1/cli/lsp/documents.rs#L1437-L1448. I tried moving it to a blocking thread with a cancellation token triggered on shutdown, but something else was still blocking the shutdown notification from being received.

Either way, It would be good to increase this timeout for projects with large dep graphs like fresh.

cc @marvinhagemeister

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -141,7 +141,8 @@ export function startLanguageServer(
       extensionContext.testController = undefined;
       extensionContext.statusBar.refresh(extensionContext);
       vscode.commands.executeCommand("setContext", ENABLEMENT_FLAG, false);
-      await client.stop(1e4);
+      const timeoutMs = 10_000;
+      await client.stop(timeoutMs);
     }
 
     // Start a new language server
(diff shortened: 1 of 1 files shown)


bartlomieju:
@nayeemrmn are you sure that it's something else blocking when that code is moved to a separate thread? Maybe we should check if the request was cancelled in-between loop iterations? Pinging @dsherret for posterity

client/src/commands.ts
```diff
@@ -141,7 +141,7 @@ export function startLanguageServer(
       extensionContext.testController = undefined;
       extensionContext.statusBar.refresh(extensionContext);
       vscode.commands.executeCommand("setContext", ENABLEMENT_FLAG, false);
-      await client.stop();
+      await client.stop(1e4);
```
dsherret, line 144:
```suggestion
      const timeoutMs = 10_000;
      await client.stop(timeoutMs);
```

Nitpick: recommend extracting a descriptive variable and not using `e` because it's less familiar.

dsherret (approved):
LGTM. Would be nice if we could get cancellation working eventually in the doc analyzer.

PR #962, opened 2023-10-08
refactor: move "deno.reloadImportRegistries" command to the server

nayeemrmn:
This will let other editors use it for free. I'll actually add the command on the LSP after this lands (https://github.com/denoland/deno/pull/20823), I want to include the date for this change in a comment there. Backwards-compatible.


bartlomieju (approved):
LGTM

PR #964, opened 2023-10-08
fix: update tasks sidebar on config change notification

### LINKED ISSUE #963: Task sidebar doesn't auto-update until before the next config file save
The task sidebar added in #958 refreshes on `deno.json` save by querying the server, but this happens too early. The server hasn't updated it's `deno.json` store yet. We need a custom server-to-client notification for config file changes.

nayeemrmn:
Fixes #963. Coupled with https://github.com/denoland/deno/pull/20827. The tasks sidebar will now only work with Deno 1.37.2 or whichever version that PR lands for, since it depends on the `sourceUri` being included in task definitions.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -396,11 +396,14 @@ export async function activate(
       treeDataProvider.refresh();
     }
   }));
-  context.subscriptions.push(vscode.workspace.onDidSaveTextDocument((event) => {
-    if (event.uri.fsPath.match(/\/deno\.jsonc?$/)) {
-      treeDataProvider.refresh();
-    }
-  }));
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM. Curious; does this panel list both "tasks" from "deno.json" file as well as "scripts" from "package.json"? (That's what `deno task` does internally, adding a comment that certain script is from package.json).

nayeemrmn:
> does this panel list both "tasks" from "deno.json" file as well as "scripts" from "package.json"?

It doesn't apparently, we would patch that in main repo by including it in the `deno/task` response

PR #981, opened 2023-10-19
chore: deprecate "deno.*" mirror settings

nayeemrmn:
- Deprecate our mirrors of built-in settings that we now support under `typescript.*` and `javascript.*`.
- Mark some settings as `resource`-scoped so they can vary between workspace folders. Depends on https://github.com/denoland/deno/pull/20937.
- I realised `deno.codeLens.implementations`, `deno.codeLens.references` and `deno.codeLens.referencesAllFunctions` are also mirrors of built-in settings. Start sending over their original counterparts so we can deprecate those too.


bartlomieju (approved):
LGTM

PR #1007, opened 2023-12-12
feat: "deno.internalInspect" setting

nayeemrmn:
Coupled with https://github.com/denoland/deno/pull/21482.

### DIFF
--- a/client/src/shared_types.d.ts
+++ b/client/src/shared_types.d.ts
@@ -102,6 +102,7 @@ export interface Settings {
   /** A flag that enables additional internal debug information to be printed
    * to the _Deno Language Server_ output. */
   internalDebug: boolean;
+  internalInspect: boolean;
   /** Determine if the extension should be providing linting diagnostics. */
   lint: boolean;
   logFile: boolean;
--- a/client/src/shared_types.d.ts
+++ b/client/src/shared_types.d.ts
@@ -102,7 +102,7 @@ export interface Settings {
   /** A flag that enables additional internal debug information to be printed
    * to the _Deno Language Server_ output. */
   internalDebug: boolean;
-  internalInspect: boolean;
+  internalInspect: boolean | string;
   /** Determine if the extension should be providing linting diagnostics. */
   lint: boolean;
   logFile: boolean;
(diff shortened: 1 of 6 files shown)


bartlomieju (commented):
Sorry, I forgot to mention - maybe we need to actually provide an address for this setting? When I was using that inspector yesterday I was getting panics in the editor because previous LSP instance was already bound. WDYT?

nayeemrmn:
> Sorry, I forgot to mention - maybe we need to actually provide an address for this setting? When I was using that inspector yesterday I was getting panics in the editor because previous LSP instance was already bound. WDYT?

Isn't there some allocation for those addresses in the CLI? Like if you launch two inspected process at once: 
image

bartlomieju:
> > Sorry, I forgot to mention - maybe we need to actually provide an address for this setting? When I was using that inspector yesterday I was getting panics in the editor because previous LSP instance was already bound. WDYT?
> 
> Isn't there some allocation for those addresses in the CLI? Like if you launch two inspected process at once: image

Nope - for regular `deno run` you can provide the address in `--inspect` flags, for the LSP for now I hardcoded the address to `127.0.0.1:10101`

nayeemrmn:
But in the screenshot I provided they have automatically hosted themselves on different pathnames on the same host/port, and there's no error. Though only the first one seems to be listed on devtools.

bartlomieju:
> But in the screenshot I provided they have automatically hosted themselves on different pathnames on the same host/port, and there's no error. Though only the first one seems to be listed on devtools.

Oh I see what you mean. But if you have a Deno instance running with an inspector, then the LSP would try to bind to that port for inspector server. And that would still cause it to fail, because something is bound to that port.

client/src/shared_types.d.ts
```diff
@@ -102,6 +102,7 @@ export interface Settings {
   /** A flag that enables additional internal debug information to be printed
    * to the _Deno Language Server_ output. */
   internalDebug: boolean;
+  internalInspect: boolean;
```
bartlomieju, line 105: This doesn't seem to be correct? Shouldn't it be `boolean | string`?

bartlomieju (approved):

Issue #908, opened 2023-09-05
Empty `deno.enablePaths` should disable for all paths


nayeemrmn:
Currently we default `deno.enablePaths` to `[]`, and throughout the extension and LSP we specially handle the empty case to mean `deno.enablePaths` was unspecified. That's not the way to handle an opt-in allowlist. It should be a nullable field.

Use cases:
- If a user has a single element in their `deno.enablePaths` and they want to temporarily disable that path, they will delete it and unexpectedly enable everything.
- If `deno.enablePaths` is determined programmatically, this is at best an annoying quirk to handle and worse if they don't know about it.

https://github.com/denoland/vscode_deno/blob/d60350a5edb25eef8b8872a68bd428eb3aa57ef3/package.json#L146-L151

PR #1028, opened 2024-01-15
chore: default "deno.enablePaths" to null

nayeemrmn:
Towards #908. For the last step we have to wait some more version cycles and then remove this code from the server: https://github.com/denoland/deno/blob/v1.39.4/cli/lsp/config.rs#L793-L795. No changes are required to the compat table since it already excludes whatever version pairs this would affect.


dsherret (approved):
LGTM!

PR #1031, opened 2024-01-20
refactor: clean up enable settings on plugin interface

nayeemrmn:
Align enable-checks with how they are done in the server in preparation for multi-deno.json support.

This exposed lots of cruft in the way we were handling workspace settings, e.g. still storing and refreshing per-document settings, our explicit type definitions for them being redundant and other things.

Also remove compat paths for LSP versions <= 1.37.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,8 @@ The extension provides several commands:
   > provide a quick fix to fetch and cache those dependencies, which invokes
   > this command for you.
 
-- _Deno: Enable_ - will enabled Deno on the current workspace. Alternatively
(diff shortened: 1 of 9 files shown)


bartlomieju (approved):
LGTM, very nice cleanup. Before landing could you check what will happen if a user tries to use this version of extension with Deno older than v1.37?

nayeemrmn:
Version <1.37.0 already crash on init as of #1028, nothing to do with this PR. This PR affects 1.37.0 exactly. It starts fine but some palette commands no longer work i.e. reload import registries, cache current module.

PR #1034, opened 2024-01-24
feat: accept notifications for subdirectory deno.json files

nayeemrmn:
This completes the client-side changes for https://github.com/denoland/deno/pull/21029.

Also fixes disposable handling for resources created with the language client. Some were only being disposed when the extension deactivated, now their lifetimes are properly tied to the client. There were no bugs from this as far as I can tell. Just dangling event handlers.


bartlomieju (approved):
LGTM 👍

PR #1057, opened 2024-01-30
chore: rename "deno/task" request to "deno.taskDefinitions"

nayeemrmn:
Towards TODO mentioned in https://github.com/denoland/deno/blob/v1.40.2/cli/lsp/mod.rs#L66-L68. This bumps the minimum required CLI version from 1.37.1 to 1.37.2. I also inverted the compat table, since the CLI version is more likely to be fixed for the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,14 +221,15 @@ extension has the following configuration options:
 
 ## Compatibility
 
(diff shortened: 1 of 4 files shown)


dsherret (approved):
LGTM

PR #1060, opened 2024-01-31
feat: Show a modal dialog after workspace init

irbull:
Previously, the user was prompted with a few questions about their workspace configuration, making it clear that something was happening. Now that there is no more user input required, it is not clear that the extension is doing anything.

This change-set shows a show a modal info box to the user after the Deno workspace is initialized.

In a future commit, this should be replaced with a message to the status bar.


nayeemrmn (approved):
LGTM. Thanks!

PR #1064, opened 2024-02-13
doc: mention other settings that affect this extension

king8fisher:
I bumped into this and spent an hour hunting why there was no suggestion popping up for refactoring. I'm mostly hoping this to remind me in the future where to look.

Also hoping maybe all the rest of the `client/src/shared_types.d.ts` might be eventually also documented for people to be aware of how the vscode_deno is affected by those.

Assuming that this repo is the source of truth for documentation, and https://docs.deno.com/runtime/manual/references/vscode_deno/ derives from here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,11 @@ extension has the following configuration options:
   lenses. This is the equivalent to using `--unstable` on the command line.
   _boolean, default `false`_
 
+There are other settings that will affect whether this extension will be triggered to perform certain actions:
+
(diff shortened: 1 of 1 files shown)


nayeemrmn:
We aim to support all of the options in `typescript`/`javascript` that are applicable, and there are more than just this one. I'm not sure it's worth explicitly documenting them. The default for this option is `"prompt"` so I don't think most users would hit this. Thanks for the PR anyway 🙏

PR #1083, opened 2024-03-05
fix: Explicitly allow null in deno.enable

irbull:
The default value for `Deno.enable` was `null` which is not valid. This commit removes the `null` value and makes the default value `false`.

This change-set also updates the formatting of the `internalInspect`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,10 @@
       "title": "Deno",
       "properties": {
         "deno.enable": {
-          "type": "boolean",
+          "type": [
+            "boolean",
+            "null"
+          ],
           "default": null,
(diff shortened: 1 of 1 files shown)


nayeemrmn:
The default value is intentionally `null`, that's the value which says to defer to `deno.json` detection. Users hitting a bug from that are (usually unknowingly) using an outdated version of Deno.

irbull:
🤔 Hum, I guess it's a bit too late to make that an emum instead? The VSCode tooling reports a warning for `null`.

Screenshot 2024-03-05 at 2 56 35 PM

nayeemrmn:
> The VSCode tooling reports a warning for `null`.

Ah, that's because we do it like
```js
{
  "deno.enable": {
    "type": "boolean",
    "default": null,
  }
}
```
which I guess I had seen in other options when adding this. I guess we're encouraging users to omit it instead of using an explicit `null`? Anyway a PR would be welcome to change this and similar options to `"type": ["boolean", null]`.

irbull:
+1, yep, PR incoming

irbull:
According to the VSCode inline tooltips, it should be an array of strings representing the subset, so `["boolean", "null"]` not `["boolean", null]`. [1]. Although the later seems to work too

[1] https://github.com/microsoft/vscode-json-languageservice/blob/main/src/services/configuration.ts#L492

nayeemrmn (approved):
LGTM. Thanks!

PR #1095, opened 2024-03-27
docs: correct the deno.unstable description

irbull:
the `deno.unstable` flag is only used to control test execution. This change-set updates the description to reflect that.


nayeemrmn (approved):
Thanks!

PR #1111, opened 2024-05-07
fix: handle nested config files in tasks sidebar

nayeemrmn:
Use the correct workspace folder and `-c subdir/deno.json` when executing subdirectory `deno.json` tasks via the sidebar. Also dedup some code.

### DIFF
--- a/client/src/tasks_sidebar.ts
+++ b/client/src/tasks_sidebar.ts
@@ -141,6 +141,7 @@ class DenoTaskProvider implements TaskProvider {
         workspaceFolders.reverse();
         const workspaceFolder = workspaceFolders.find((f) =>
           // NOTE: skipEncoding in Uri.toString(), read more at https://github.com/microsoft/vscode/commit/65cb3397673b922c1b6759d145a3a183feb3ee5d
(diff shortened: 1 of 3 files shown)


client/src/tasks_sidebar.ts
```diff
@@ -144,47 +124,45 @@ class DenoTaskProvider implements TaskProvider {
...
+    const tasks = [];
+    try {
+      const configTasks = await client.sendRequest(taskReq);
+      for (const configTask of configTasks ?? []) {
+        const workspaceFolders = Array.from(
+          workspace.workspaceFolders ?? [],
+        );
+        workspaceFolders.reverse();
+        const workspaceFolder = workspaceFolders.find((f) =>
+          // NOTE: skipEncoding in Uri.toString(), read more at https://github.com/microsoft/vscode/commit/65cb3397673b922c1b6759d145a3a183feb3ee5d
```
dsherret, line 143: I realize this is previous code, but it might be good if we explained why `skipEncoding` here.

dsherret (approved):
LGTM

Issue #1153, opened 2024-08-31
Support for `unstable-css` as CSS formatter in VSCode


isAdrisal:
With Deno 1.46, it's now possible to format CSS styles with Deno:

```
deno fmt --unstable-css styles.css
```

This works well, but it doesn't seem supported by the VSCode extension, which means it can't be used to format style on save. Setting `denoland.vscode-deno` as the default CSS formatter results in this error:

`Extension 'Deno' is configured as formatter but it cannot format 'CSS'-files`

**Describe the solution you'd like**

Support for the new `unstable-css` formatter in the extension – or guidance on how to get it working if it can already.

bartlomieju:
@nayeemrmn could you handle that?

nayeemrmn:
Closed by https://github.com/denoland/deno/pull/25353 and #1154.

isAdrisal:
Hey @nayeemrmn, I think I'm missing a step to get this working.

I have v3.39.0 of the VSCode extension installed, and this in my VSCode settings:

```
{
  "deno.enable": true,
  "deno.lint": true,
  "deno.unstable": true,
  "editor.formatOnSave": true,
  "[css]": {
      "editor.defaultFormatter": "denoland.vscode-deno"
  }
}
```

I'm no longer seeing the `cannot format 'CSS'-files` anymore, but no formatting happens after save of a CSS file. Format on save for js/ts/tsx is working as normal.

Any help would be appreciated :)

isAdrisal:
@bartlomieju Makes sense, thanks!

PR #1154, opened 2024-09-02
chore: sync css documents with the lsp

nayeemrmn:
Client-side changes for #1153. Alongside https://github.com/denoland/deno/pull/25353.


bartlomieju (approved):

bartlomieju (changes requested):
Actually, we should also add HTML and YAML support. We can leave various "component" files for a follow up (please open a tracking issue though).

bartlomieju (approved):

PR #1166, opened 2024-09-23
feat: helper prompts for tsconfig.json compiler options

nayeemrmn:
Related: https://github.com/denoland/deno/issues/25607.

image

image

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -418,7 +418,7 @@ async function maybeShowTsConfigPrompt(
     }
     const [scopeFolderPath, scopeFolder] = scopeFolderEntry;
     selection = await vscode.window.showInformationMessage(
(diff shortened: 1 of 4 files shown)


client/src/commands.ts
```diff
@@ -518,3 +682,41 @@ function isDenoDisabledCompletely(): boolean {
...
+  "rootDirs",
+  "skipLibCheck",
+  "sourceMap",
+  "sourceRoot",
+  "stripInternal",
+  "tsBuildInfoFile",
+  "typeRoots",
+  "watch",
+  "watchDirectory",
+  "watchFile",
```
dsherret, line 721: I don't really like the idea of having another list of compiler options here, but I can't think of anything else and this is only used for the conversion.

dsherret (approved):
LGTM

Issue #1190, opened 2024-10-23
Code generated by `deno init` gives an error in VS Code


mike-lischke:
deno 2.0.2 (stable, release, aarch64-apple-darwin)
v8 12.9.202.13-rusty
typescript 5.6.2

The generated code is as expected:

```ts
export function add(a: number, b: number): number {
  return a + b;
}

// Learn more at https://docs.deno.com/runtime/manual/examples/module_metadata#concepts
if (import.meta.main) {
  console.log("Add 2 + 3 =", add(2, 3));
}
```

however, this gives me the error:

> The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node16', or 'nodenext'.ts(1343)

![Bildschirmfoto 2024-10-23 um 17 19 59](https://github.com/user-attachments/assets/178565ca-97cd-4932-be9f-1bed66e743c0)

All I did was to create a new folder, opened that in VS Code and ran `deno init`. Then I opened main.ts and got this. Not a good first impression, even though `deno main.ts` (or `deno run main.ts` for that matter) run fine.

Further info: I have tsc 5.2.2 installed globally on my box and .vscode/settings.json just contains:

```json
{
    "deno.enable": true
}
```

nayeemrmn:
Check your user settings for `"typescript.tsserver.experimental.enableProjectDiagnostics": true`, and disable it. Those errors are coming from the built-in TS server, and our plugin can't supress errors caused by that setting.

Please verify if that was the problem.

mike-lischke:
That was it @nayeemrmn! Removing this key from the settings.json file made the TS diagnostics go away.

mike-lischke:
Thanks for analyzing this problem with me. It's obviously a user-generated problem. Would be good if this setting could be checked by the extension and a warning be shown to inform the user about possible conflicts. Thanks!

nayeemrmn:
@mike-lischke No problem! I've already opened a PR for this: #1191.

Issue #1184, opened 2024-10-16
An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled


khanakia:
**Describe the bug**
I just did a fresh install and opened the project in vscode and getting weird typescript errors

I have enabled Deno and followed the steps here https://docs.deno.com/runtime/reference/vscode/#troubleshooting still no luck
**To Reproduce**

1. deno run -A -r https://fresh.deno.dev
2. cd fresh-project
3. deno task start

**Expected behavior**

**Screenshots**

![Screenshot 2024-10-16 at 11 25 30 PM](https://github.com/user-attachments/assets/fa2376b3-b10b-4dd8-9e8f-38def77b499e)

![Screenshot 2024-10-16 at 11 25 43 PM](https://github.com/user-attachments/assets/31253ef3-ae59-4df0-90ab-84ffacc77377)
![Screenshot 2024-10-16 at 11 25 57 PM](https://github.com/user-attachments/assets/c4293c55-b4a9-43e2-ba9c-78565ba5e4a2)

**Versions**

vscode: 1.94.2 (Universal) deno: 2 extension: lastest

nayeemrmn:
Check your user settings for `"typescript.tsserver.experimental.enableProjectDiagnostics": true`, and disable it. Those errors are coming from the built-in TS server, and our plugin can't supress errors caused by that setting.

Please verify if that was the problem.

khanakia:
thanks this fixed the issue

Issue #1098, opened 2024-03-31
Nothing works :(


aspizu:
Keep getting typescript errors, even tried "disabling" the builtin typescript extension which apparently doesn't actually disable the extension at all. Trying to contribute to a deno fresh project, the repo contains a `.vscode/settings.json` with `deno.enable: true`.

I get errors such as:
```json
[{
	"resource": "/home/aspizu/Documents/Projects/Contrib/jsr/frontend/routes/package/versions.tsx",
	"owner": "_generated_diagnostic_collection_name_#0",
	"code": "2318",
	"severity": 8,
	"message": "Cannot find global type 'Promise'.",
	"source": "typescript",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 2
}]
```

```
Deno Version: 3.36.0
VSCode Version: 1.87.2 (AUR visual-studio-code-bin)
```

nayeemrmn:
Check if you have `typescript.tsserver.experimental.enableProjectDiagnostics` enabled in your **user** settings. If so you'll have to turn that off.

aspizu:
I don't have that setting in my user settings.

codinronan:
@aspizu what was your resolution? I have this same issue now. When I use Command + Shift + P and do "Deno language server status" NONE of the settings match the ones in my `.vscode/settings.json` NOR `deno.jsonc` files. They're all apparently the defaults so it doesn't even see my deno configs at all. This has broken imports, type checking, formatting, the lot.

What was your fix?

nayeemrmn:
@codinronan Are those relative to your root directory? Please show a screenshot or paste of your explorer view

codinronan:
@nayeemrmn I found the answer in another issue, https://github.com/denoland/vscode_deno/issues/1099 about deno.jsonc now needing to be specified. Once I did that it worked!

(Fwiw yes they're relative to the root directory).

aspizu:
The extension for [Civet](https://civet.dev/) creates this issue, which needs to be disabled. Should this disable that extension, or should it be up to the user?

Issue #1072, opened 2024-02-19
VSCode Cannot find name 'Deno'  ts(2304)


nah-sonrisa:
**Describe the bug**

![image](https://github.com/denoland/vscode_deno/assets/155512350/c692b717-e5dd-4a78-a21d-98d4cdb7e59c)

settings.json:
```
{
  "deno.enablePaths": [
    "supabase/"
  ],
  "deno.lint": true,
  "deno.unstable": true,
  "[typescript]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "deno.enable": true,
  "typescript.validate.enable": true,
  "javascript.validate.enable": true,
  "deno.path": "/opt/homebrew/bin/deno",
  "typescript.tsserver.experimental.enableProjectDiagnostics": true
}
```

Output
```
Starting Deno language server...
  version: 1.40.5 (release, aarch64-apple-darwin)
  executable: /opt/homebrew/bin/deno
Connected to "Visual Studio Code" 1.86.2
Enabling import suggestions for: https://deno.land
Server ready.
Starting Deno language server...
  version: 1.40.5 (release, aarch64-apple-darwin)
  executable: /opt/homebrew/bin/deno
Connected to "Visual Studio Code" 1.86.2
Enabling import suggestions for: https://deno.land
Server ready.
Enabling import suggestions for: https://deno.land
```

I have tried to set "typescript.tsserver.experimental.enableProjectDiagnostics": true and restart, the error still exists. 

Although when I run `deno check functions/index.ts` in terminal it doesn't error, so I think it's still an issue with vscode settings.

**Versions**

vscode: 1.86.2 deno: 1.40.5 extension: v3.33.3

nayeemrmn:
For me this only happens with `"typescript.tsserver.experimental.enableProjectDiagnostics": true`, which is an issue we haven't been able to work around. Report back here if it still happens with that setting off.

PR #1191, opened 2024-10-23
feat: disable "enableProjectDiagnostics" on enable

nayeemrmn:
The diagnostics from this setting cannot be suppressed by our plugin: https://github.com/denoland/vscode_deno/issues/437#issuecomment-1720393193. When users see diagnostics from both `ts` and `deno-ts`, this has always been the cause. So disable it for the workspace if it's enabled there or in user settings.

Non-exhaustive list of reports:
https://github.com/denoland/vscode_deno/issues/66#issuecomment-1316154839
#466
https://github.com/denoland/vscode_deno/issues/668#issuecomment-1159481069
#715
#1072
#1098
#1184
#1190

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -655,6 +655,25 @@ export function enable(
     const config = vscode.workspace.getConfiguration(EXTENSION_NS);
     await config.update("enable", true);
     vscode.window.showInformationMessage("Deno workspace initialized.");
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
Great, thanks

alumni:
I think there might be a regression on this.

Issue #1211, opened 2024-11-22
Unable to detect tasks with the new object notation format


hugoalh:
**Describe the bug**

Deno v2.1.0 introduced the new object notation format for the tasks (see https://github.com/denoland/deno/pull/26886 or https://deno.com/blog/v2.1#turbocharged-deno-task), however this is not yet supported by the extension.

**Versions**

vscode: 1.95.3 deno: 2.1.1 extension: 3.42.0

PR #1214, opened 2024-11-26
fix: support task object notation in sidebar

nayeemrmn:
Coupled with https://github.com/denoland/deno/pull/27076. Fixes part of #1211.


bartlomieju (approved):
Great!

PR #1243, opened 2025-01-21
Run tests with file watcher

### LINKED ISSUE #1132: Watch mode for "Testing code lens"
**Is your feature request related to a problem? Please describe.**

The Deno extension has a [very nice feature](https://docs.deno.com/runtime/manual/references/vscode_deno/#testing-code-lens) that allows running tests from the VSCode interface. However, this involves manually clicking on "▶ Run Test" each time.

Deno CLI has `deno test --watch`. It'd be convenient to have that via the extension as well.

**Describe the solution you'd like**

Add the ability to automatically re-run tests in the sidebar when the test file is saved. It would ideally be next to the "Run tests" button, as a "Run tests and watch" option.

image

stefnotch:
I wanted a Deno testing experience that's closer to Vitest's UI or Wallaby.js. As in
- See the test results inline in the IDE
- Re-run the tests and update the IDE's UI when saving 

This now works 🎉 
https://github.com/user-attachments/assets/b283b1e1-528f-47d5-b436-86c546d0d3c7

(The execution speed is very slow, because I have not compiled Deno in release mode)

To do so, I had to
- Bump the required VSCode version. This lets us take advantage of VSCode's "continuous run".
- Update the LSP to support these features. https://github.com/denoland/deno/pull/27762
- Recreate the TestRun object when the tests are being restarted. This is because VSCode does not support changing the status of a test from failure back to success.
- Learn about a VSCode peculiarity https://github.com/microsoft/vscode/issues/172205

Fix #1132

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "deno"
   ],
   "engines": {
-    "vscode": "^1.60.0"
+    "vscode": "^1.77.0"
   },
   "activationEvents": [
     "onLanguage:typescript",
@@ -257,7 +257,9 @@
           "markdownDescription": "Additional environment variables to pass to Deno processes. Overrides the user's env and `deno.envFile`. These will be overridden by more specific settings such as `deno.future` for `DENO_FUTURE`, and invariables like `NO_COLOR=1`.",
           "scope": "window",
           "examples": [
-            { "HTTP_PROXY": "http://localhost:8080" }
+            {
+              "HTTP_PROXY": "http://localhost:8080"
+            }
           ]
         },
         "deno.envFile": {
@@ -479,7 +481,11 @@
           "scope": "resource"
         },
         "deno.trace.server": {
-          "enum": ["messages", "off", "verbose"],
+          "enum": [
+            "messages",
+            "off",
+            "verbose"
+          ],
           "default": "off",
           "markdownDescription": "Traces the communication between VS Code and the Deno Language Server.",
           "scope": "window"
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -16,11 +16,11 @@
       },
       "devDependencies": {
         "@types/semver": "7.3.9",
-        "@types/vscode": "1.67.0",
+        "@types/vscode": "1.77.0",
         "vscode-test": "^1.6.1"
       },
       "engines": {
-        "vscode": "^1.60.0"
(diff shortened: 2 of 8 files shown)


stefnotch:
Related PR with questions has been opened https://github.com/denoland/deno/pull/27762

package.json
```diff
@@ -29,7 +29,7 @@
     "deno"
   ],
   "engines": {
-    "vscode": "^1.60.0"
+    "vscode": "^1.77.0"
```
nayeemrmn, line 32: Does this make a difference to the currently resolved version and is it essential? I'd like to avoid it because the TS plugin API communication has been brittle in the past

stefnotch: Yes, this bit is essential. It's what is required to get `request.continuous` and the final argument in
```
new vscode.TestRunRequest(
              runData.request.include,
              runData.request.exclude,
              runData.request.profile,
              true,
            ),
```

Without that, we don't have the typings for the continuous run feature in vscode.

stefnotch:
Thank you for merging this!

Would you have any suggestions regarding the Deno LSP PR that implements the internal functionality for this?
https://github.com/denoland/deno/pull/27762

One of my biggest questions was: There's fair bit of code duplication im Deno. The same sort of "run normal" vs "run in watch mode" logic is duplicated here:
run_tests
https://github.com/denoland/deno/blob/57dd66ec3dae2a0fe376f6a43c476dfade421c84/cli/tools/test/mod.rs#L1555
run_tests_with_watch
https://github.com/denoland/deno/blob/57dd66ec3dae2a0fe376f6a43c476dfade421c84/cli/tools/test/mod.rs#L1647

exec
https://github.com/denoland/deno/blob/57dd66ec3dae2a0fe376f6a43c476dfade421c84/cli/lsp/testing/execution.rs#L216

run_benchmarks
https://github.com/denoland/deno/blob/57dd66ec3dae2a0fe376f6a43c476dfade421c84/cli/tools/bench/mod.rs#L426
run_benchmarks_with_watch
https://github.com/denoland/deno/blob/57dd66ec3dae2a0fe376f6a43c476dfade421c84/cli/tools/bench/mod.rs#L492


Now I did just "exec_with_watch" to the LSP and call it a day, but I feel like there might be a better approach. Any recommendations?

PR #1258, opened 2025-02-25
fix: move extended language formatting to client

### LINKED ISSUE #1228: Vue component highlighting not working
**Describe the bug**
Since installing Deno extension, Vue's component highlighting has stopped working and is displayed in the same color as default HTML5 tags. Intellisense functions properly, and I have confirmed that highlighting works correctly when `Deno` is uninstalled.

**To Reproduce**

1. Install [`Vue - Official`](https://marketplace.visualstudio.com/items?itemName=Vue.volar) extension
2. Install `Deno` extension
3. Create a User-Defined component in Vue.js SFC file

**Expected behavior**
The custom component highlighting in Vue needs to work properly.

**Screenshots**

image
Before installing Deno

image
After installing Deno

**Versions**

vscode: 1.93.0 deno: 2.1.4 extension: 3.43.1

Vue - Official: I have confirmed the same symptom in all versions from 2.2.0 to 1.8.27.

nayeemrmn:
Fixes #1228.

Ref https://github.com/denoland/deno/pull/28262.


bartlomieju (commented):
Won't this be slower than formatting via the LSP?

nayeemrmn:
> Won't this be slower than formatting via the LSP?

Well, it doesn't run in the language server lock. Plus it only applies to these languages that have no business going through the LSP. This is probably the correct way of handling them.

bartlomieju (approved):

PR #1271, opened 2025-03-09
feat(enablePaths): glob support

### LINKED ISSUE #834: deno.enablePaths: pattern support
**Is your feature request related to a problem? Please describe.**

I work in a lot of projects that uses Node.js and TypeScript. I often want to write scripts in sub directories using Deno though. My list for `deno.enablePaths` is becoming super long and it is annoying each time I create a new sub directory and I have to manually add it to the list.

**Describe the solution you'd like**

It turns out that most of my sub directories follow the same pattern. I'd love to define a regular expression or glob pattern instead of explicitly listing and maintaining the list of every sub directory.

scarf005:
resolves #834

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,10 +11,12 @@
       "dependencies": {
         "dotenv": "^16.4.5",
         "jsonc-parser": "^3.2.0",
+        "picomatch": "^4.0.2",
         "semver": "7.5.2",
         "vscode-languageclient": "^8.0.1"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


nayeemrmn:
This needs to be implemented server-side as well

Mrashes:
@scarf005 is server-side implementation in progress?

scarf005:
nope, could i get pointers on where to look?

Mrashes:
I'm pretty new to the code base, @nayeemrmn - any direction?

Issue #1257, opened 2025-02-22
Deno Language Server client: couldn't create connection to server. Error: spawn EINVAL


cvbge:
**Describe the bug**

I'm getting error after `Deno: Initialize Workspace Configuration`:

```
[Error - 8:23:27 PM] Deno Language Server client: couldn't create connection to server.
Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at Object.spawn (node:child_process:796:9)
    at c:\Users\Piotr\.vscode\extensions\denoland.vscode-deno-3.43.3\client\dist\main.js:18375:40
    at LanguageClient2.createConnection (c:\Users\Piotr\.vscode\extensions\denoland.vscode-deno-3.43.3\client\dist\main.js:17485:28)
    at LanguageClient2.start (c:\Users\Piotr\.vscode\extensions\denoland.vscode-deno-3.43.3\client\dist\main.js:17101:30)
    at c:\Users\Piotr\.vscode\extensions\denoland.vscode-deno-3.43.3\client\dist\main.js:21138:5
    at activate (c:\Users\Piotr\.vscode\extensions\denoland.vscode-deno-3.43.3\client\dist\main.js:22146:3)
    at aw.n (file:///c:/Users/Piotr/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:112:4447)
    at aw.m (file:///c:/Users/Piotr/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:112:4410)
    at aw.l (file:///c:/Users/Piotr/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:112:3866)
```

**To Reproduce**

1. Install deno with `npm install -g deno`.
2. Create project with `deno init`.
3. Open it with `code`.
4. Install `deno` extension.
5. Run `Deno: Initialize Workspace Configuration`.

The error appears in Output view.

**Versions**

vscode: 
```
Version: 1.97.2 (user setup)
Commit: e54c774e0add60467559eb0d1e229c6452cf8447
Date: 2025-02-12T23:20:35.343Z
Electron: 32.2.7
ElectronBuildId: 10982180
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
[...]

PR #1273, opened 2025-03-12
fix(client): set `{ shell: true }` in server options for "deno.{bat,cmd}"

BinToss:
Fixes "Deno Language Server client: couldn't create connection to server. Error: spawn EINVAL" #1257

I've debugged it locally to ensure it works on my Windows PC.

---

`vscode_deno` is currently broken on Win32 VSCode 1.92 (July 2024) and later. `spawn`'s/`spawnSync`'s EINVAL error was introduced in the Node.js 18.x, 20.x, 21.x releases lines to mitigate CVE-2024-27980. The fix was implemented at nodejs/node@69ffc6d50d.

Running a .bat or .cmd via spawn or spawnSync _without assigning the shell option_ is prohibited in VSCode 1.92 and later.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -32,7 +32,11 @@ import { DenoServerInfo } from "./server_info";
 import * as dotenv from "dotenv";
 import * as vscode from "vscode";
 import { LanguageClient, ServerOptions } from "vscode-languageclient/node";
-import type { Location, Position } from "vscode-languageclient/node";
+import type {
+  Executable,
+  Location,
+  Position,
+} from "vscode-languageclient/node";
 import { getWorkspacesEnabledInfo, isPathEnabled } from "./enable";
 import { denoUpgradePromptAndExecute } from "./upgrade";
 import * as fs from "fs";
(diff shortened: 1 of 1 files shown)


client/src/commands.ts
```diff
@@ -164,19 +168,21 @@ export function startLanguageServer(
...
+      } as Executable,
       debug: {
         command,
         // disabled for now, as this gets super chatty during development
         // args: ["lsp", "-L", "debug"],
         args: ["lsp"],
-        options: { env },
-      },
+        options: { env, shell },
+      } as Executable,
```
nayeemrmn, line 185: Why are the casts necessary? `as Executable`.

nayeemrmn:
Thanks for the fix @BinToss! If you can sign the CLA we'll land this

client/src/commands.ts
BinToss: It was just for typed intellisense so I knew what data type was expected. Without the cast, the structs weren't typed appropriately.

If necessary, I can revert it or use narrowing and type extraction to extract the correct type from the ServerOptions union.

BinToss:
Done

PR #1278, opened 2025-04-01
feat: sync vscode-userdata schemed documents with the lsp

### LINKED ISSUE #1206: Unable to format vscode user profile level `settings.json` file
**Describe the bug**

```sh
Extension 'Deno' is configured as formatter but it cannot format 'JSON with Comments'-files
```

The [`denoland.vscode-deno`](https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno) extension is ONLY cannot format the user profile level `settings.json` file.

But it can format the workspace level `settings.json` file or the `*.jsonc` file within the workspace.

This is very strange but I think it might be solved in vscode extensions since [`Prettier - Code formatter`](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension works in all files mentioned above.

This is actually consistent with #840, but I think it defines the scope of the issue more clearly here, and the possibility of a fix.

I've also tested adding `"editor.formatOnSaveMode": "file"` but it has no effect.

Like said in the [comment](https://github.com/denoland/vscode_deno/issues/840#issuecomment-1520950590), if you drag that `settings.json` as a file into vscode to open it works, but not when opening the file directly from entrances within vscode.

**NOTE** that the Prettier extension works in both cases, so there must be a way around it.

Not supporting formatting of user profile level `settings.json` file is very inconvenient, for example I cannot set "denoland.vscode-deno" to user level "editor.defaultFormatter" because it does not work in the file itself.
Instead I have to configure "denoland.vscode-deno" individually in each workspace.

I hope someone can investigate and resolve this humble little issue. And I'm perfectly willing to set this to my user level "editor.defaultFormatter" once this issue is resolved.

**Versions**

vscode: `1.95.2` deno: `2.0.6` extension: `3.42.0`

nayeemrmn:
Fixes #1206. Coupled with https://github.com/denoland/deno/pull/28706.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -143,6 +143,7 @@ export async function activate(
       { scheme: "deno", language: "astro" },
       { scheme: "deno", language: "vento" },
       { scheme: "deno", language: "nunjucks" },
+      { scheme: "vscode-userdata", language: "jsonc" },
       { notebook: "*", language: "javascript" },
       { notebook: "*", language: "javascriptreact" },
       { notebook: "*", language: "typescript" },


dsherret (approved):
LGTM

Issue #1267, opened 2025-03-04
support Test Explorer - Colored Output


pmunin:
I have the following test:
```ts
Deno.test("output should be colored!", (t)=>{
  console.debug(`%cTerm: ${Deno.env.get('TERM')}. stdout.isTerminal: ${(Deno.stdout.isTerminal())}. noColor:${Deno.noColor}`, "color: red");
})
```
When I run it via code lense - It opens the terminal and show output in red color:

![Image](https://github.com/user-attachments/assets/06c6d06a-480a-41e7-9ca7-9c4f3a6d9a91)

However, when I run it from Test explorer - I get boring plain white test output, with no coloring:

![Image](https://github.com/user-attachments/assets/e006a74e-9d6c-4dfd-8638-97eea40540b5)

![Image](https://github.com/user-attachments/assets/39a68ef2-59cf-4717-84a9-1f73c1e15f83)

Can you please support Colored output from Test Explorer, same way it works in terminal?

stefnotch:
Temporary workaround: [`FORCE_COLOR` is supported](https://github.com/denoland/deno/pull/28490) by Deno.

You can override the `NO_COLOR` setting by going into your VSCode settings and doing

```jsonc
// .vscode/settings.json
{
  "deno.env": {
    "FORCE_COLOR": "1"
  }
}
```

Issue #1156, opened 2024-09-03
Allow `NO_COLOR` env var to be overridden by `deno.env` workspace setting


ArmaanAS:
I would like the VSCode Test Results output to be in color. I'm not able to override the `NO_COLOR` environment variable in `settings.json`, like

```jsonc
// .vscode/settings.json
{
  "deno.env": {
    "NO_COLOR": "0"
  }
}
```

This has no effect. The output remains in plain white. 

https://github.com/denoland/vscode_deno/blame/2c5964d14f44c3155da255fb7d5d63326c0761a8/client/src/commands.ts#L151

This is the line to be removed, unless there is a specific reason it must be disabled. Otherwise, move it up 6 lines, just before the so that is can be overridden by the `deno.env` setting, like

```typescript
env["NO_COLOR"] = "1";
const denoEnv = config.get<Record<string, string>>("env");
if (denoEnv) {
  Object.assign(env, denoEnv);
}
if (config.get<boolean>("future")) {
  env["DENO_FUTURE"] = "1";
}
env["DENO_V8_FLAGS"] = getV8Flags();
```

Or maybe add a `deno.color` boolean variable to the VSCode settings. Just please allow some way to enable color outputs.

stefnotch:
With [`FORCE_COLOR` now supported](https://github.com/denoland/deno/pull/28490) in Deno, you can override the `NO_COLOR` setting


```jsonc
// .vscode/settings.json
{
  "deno.env": {
    "FORCE_COLOR": "1"
  }
}
```

Proper fix hopefully coming soon to a Deno near you. :)

PR #1298, opened 2025-05-10
Enable colored output in tests

stefnotch:
I disabled the `env["NO_COLOR"]` setting for the Deno that is spawned. 

Instead, since VSCode actually handles ANSI color escapes, I *forcibly enable* the color setting. The reasoning is that it's entirely possible to start VSCode from the terminal. (e.g. when I configure my git editor to be VSCode, then VSCode gets started from the terminal).
And I do think that it would be confusing if the NO_COLOR terminal setting would affect the test output inside a GUI application. 

absolutely gorgeous colors in VSCode

Since I don't know the full impact of this change: *What should I test?*
Could this change, for example, break the Deno LSP tests?

### Investigation

As far as I can tell from the git blame, the `env["NO_COLOR"]` was added [4 years ago](https://github.com/denoland/vscode_deno/pull/454/commits/8e2119433932f13b02414e2e1ae13b90c73b653a), but I couldn't dig up further information as to why. 

### Linked issues

Fixes https://github.com/denoland/vscode_deno/issues/1156
Fixes https://github.com/denoland/vscode_deno/issues/1267

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -159,6 +159,7 @@ export function startLanguageServer(
       }
     }
     const denoEnv = config.get<Record<string, string>>("env");
+    env["FORCE_COLOR"] = "1";
     if (denoEnv) {
       Object.assign(env, denoEnv);
     }


Mrashes (commented):
Fix makes a lot of sense to me.  Not a contributor so can't approve but don't see why this can't be merged.

Issue #1303, opened 2025-05-17
Show task description from deno.json in Deno Tasks panel and command palette


narumincho:
Is your feature request related to a problem? Please describe.

In the current side panel "Deno Tasks", both the faded text (TreeItem.description) and the tooltip shown on hover (TreeItem.tooltip) always display the fixed string "task", regardless of the actual task content. This can make it difficult to distinguish between tasks at a glance.

Describe the solution you'd like

If a task has a description defined in deno.json, I would like that description to be prioritized and shown in the side panel's UI (for both TreeItem.description and TreeItem.tooltip), instead of the default "task" string. If no description is provided, the shell command should be displayed instead.

Also, in the command palette, I think it's more intuitive to show the description (if available) rather than the raw shell command, as it makes it easier to understand what each task does.

Current screenshot

![Image](https://github.com/user-attachments/assets/48a7cba5-c745-40ba-9331-a251cb440c63)

![Image](https://github.com/user-attachments/assets/f7a668a7-8cd4-4f04-8d42-61598e742273)

Screenshot after the change

![Image](https://github.com/user-attachments/assets/2954c7ff-4542-4bcd-aac0-54ce0cf3c361)

![Image](https://github.com/user-attachments/assets/e769aca6-8168-4e70-bcf4-d341c290da8a)

PR #1304, opened 2025-05-17
feat: show task description from deno.json in tasks view and command …

narumincho:
Ref: https://github.com/denoland/vscode_deno/issues/1303

### DIFF
--- a/client/src/tasks.ts
+++ b/client/src/tasks.ts
@@ -51,6 +51,7 @@ export function buildDenoConfigTask(
   name: string,
   command: string | undefined,
   sourceUri?: vscode.Uri,
+  description: string = "",
 ): vscode.Task {
   const args = [];
   if (
@@ -75,7 +76,7 @@ export function buildDenoConfigTask(
     new vscode.ProcessExecution(process, ["task", ...args]),
     ["$deno"],
   );
-  task.detail = `$ ${command}`;
+  task.detail = description || `$ ${command}`;
   return task;
 }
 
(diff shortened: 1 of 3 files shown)


client/src/tasks_sidebar.ts
```diff
@@ -98,10 +98,8 @@ class DenoTask extends TreeItem {
     this.command = commandList[defaultCommand];
     this.iconPath = new ThemeIcon("wrench");
 
-    if (this.task.definition.command) {
```
Mrashes, line 101: why remove this if check?

client/src/tasks.ts
```diff
@@ -51,6 +51,7 @@ export function buildDenoConfigTask(
   name: string,
   command: string | undefined,
   sourceUri?: vscode.Uri,
+  description: string = "",
```
Mrashes, line 54: is description able to be null?  TaskRequestResponse has it string or null, this has it as only string.  Want to make sure we are consistent.

Mrashes (commented):
looks good, thanks for making the issue and tackling it!

client/src/tasks_sidebar.ts
narumincho: I removed this if check because we're no longer using this.task.definition.command.

Even before this change, the "task" string was always being passed, so the if condition was always true anyway:
https://github.com/denoland/vscode_deno/blob/880dd486510ec37791fc51281d5af9a256366162/client/src/tasks.ts#L69

Mrashes: Appreciate the explainer

